PR #7, opened 2017-12-17
Skeleton of HTTP/2 protocol

whatyouhide:


lib/xhttp2/conn.ex
```diff
@@ -280,42 +284,26 @@ defmodule XHTTP2.Conn do
       {:ok, frame, rest} ->
         Logger.debug(fn -> "Got frame: #{inspect(frame)}" end)
 
+        if conn.headers_being_processed && not match?(continuation(), frame) do
```
ericmj: Can frames not be interleaved from other streams when headers are being processed?

whatyouhide: Nope. The spec says that if headers are split into multiple headers+continuation or push_promise+continuation frames then all those frames must come one after the other.

Quote:

> A HEADERS frame without the END_HEADERS flag set MUST be followed by a CONTINUATION frame for the same stream. A receiver MUST treat the receipt of any other type of frame or a frame on a different stream as a connection error (Section 5.4.1) of type PROTOCOL_ERROR.

Can you help me confirm the behaviour I implemented is correct?

ericmj: Hum, I wonder what the reason for splitting headers into multiple frames is then?

Yeah, seems like the correct way to do it.

whatyouhide: From what I got from the spec, there's a `max_header_list_size` which is the max size of the header block fragment in a HEADERS/PUSH_PROMISE/CONTINUATION frame, so you split over that.

whatyouhide: Errata: it's from max frame size. You shouldnt send frames bigger than max frame size so now the client automatically splits headers and data frames.

lib/xhttp2/conn.ex
```diff
@@ -0,0 +1,920 @@
...
+  end
+
+  defp handle_window_update(
+         conn,
+         window_update(stream_id: stream_id, window_size_increment: wsi),
+         responses
+       ) do
+    stream = fetch_stream!(conn, stream_id)
+
+    case conn.streams[stream_id].window_size do
```
ericmj: Are we fetching the stream twice?

lib/xhttp2/conn.ex
```diff
@@ -0,0 +1,920 @@
...
+    if stream_id != 0 do
+      debug_data = "frame #{frame} only allowed at the connection level"
+      send_connection_error!(conn, :protocol_error, debug_data)
+    end
+  end
+
+  defp send!(%__MODULE__{transport: transport, socket: socket} = conn, bytes) do
+    case transport.send(socket, bytes) do
+      :ok -> :ok
+      {:error, reason} -> throw({:xhttp, conn, reason})
```
ericmj: We might consider not throwing here. Gun avoids crashing in send because supposedly you get better error messages if you crash when receiving (and if send fails then receive will always fail).

lib/xhttp2/frame.ex
```diff
@@ -45,24 +47,37 @@ defmodule XHTTP2.Frame do
...
+  @spec set_flag(:continuation, :end_headers) :: byte()
+  @spec set_flag(:headers, :end_stream | :end_headers | :padded | :priority) :: byte()
+  def set_flag(frame_name, flag_name)
+
+  @spec flag_set?(byte(), :data, :end_stream | :padded) :: boolean()
+  @spec flag_set?(byte(), :settings, :ack) :: boolean()
+  @spec flag_set?(byte(), :push_promise, :end_headers | :padded) :: boolean()
+  @spec flag_set?(byte(), :ping, :ack) :: boolean()
+  @spec flag_set?(byte(), :continuation, :end_headers) :: boolean()
+  @spec flag_set?(byte(), :headers, :end_stream | :end_headers | :padded | :priority) :: boolean()
```
ericmj: Would it make sense to generate the specs as well?

lib/xhttp2/conn.ex
```diff
@@ -0,0 +1,920 @@
...
+  catch
+    :throw, {:xhttp, conn, error} ->
+      throw({:xhttp, conn, error, responses})
+  end
+
+  # Other frames (from any stream) MUST NOT occur between the HEADERS frame and any CONTINUATION
+  # frames that might follow.
+  # http://httpwg.org/specs/rfc7540.html#HttpSequence
+  defp assert_valid_frame!(conn, frame) do
+    if conn.headers_being_processed && not match?(continuation(), frame) do
```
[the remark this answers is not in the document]
whatyouhide: @ericmj it's not a boolean, it's `nil | {hbf_acc, ...}`.

lib/xhttp2/conn.ex
```diff
@@ -0,0 +1,920 @@
...
+
+    with {:ok, socket} <-
+           connect_and_negotiate_protocol(hostname, port, transport, transport_opts),
+         :ok <- set_inet_opts(transport, socket),
+         {:ok, conn} <- initiate_connection(transport, socket, opts) do
+      conn = %{conn | hostname: hostname, port: port, scheme: scheme}
+      {:ok, conn}
+    else
+      {:error, reason} ->
+        {:error, {:connect, reason}}
```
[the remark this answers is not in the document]
whatyouhide: @ericmj we do this [in `XHTTP1`](http://github.com/ericmj/xhttp/blob/f4d5c5e5774333b0bf8ee5f6f1fd329042683c0e/lib/xhttp1/conn.ex#L75-L76) as well. Should we remove it from both?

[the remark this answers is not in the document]
ericmj: Oh, I didn't realise that. I wonder what API is preferable?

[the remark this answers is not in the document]
whatyouhide: I think we can get rid of :connect for the reason you mentioned. Should I remove in both http1 and 2?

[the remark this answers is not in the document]
ericmj: Yes, please. 🙏

PR #25, opened 2018-03-29
XHTTP.Headers, functions for handling HTTP headers

gamache:
This PR adds the XHTTP.Headers module, containing functions for manipulating HTTP headers.  In particular, it provides convenience around case insensitivity and repeated headers, making it easier to conform to RFC 2616.

I'm not sure if you consider XHTTP the correct place for functionality like this.  Similar code is replicated ad-hoc in most HTTP tools today.  I'd prefer to see support be hoisted higher, either in Elixir's standard HTTP library (I'm assuming that's XHTTP) or higher up in the standard lib, to make it easy for end users to handle HTTP headers in a standard way.

If you agree with this approach, I can work on retrofitting XHTTP1 and XHTTP2 to use these functions internally.

### DIFF
--- a/lib/xhttp/headers.ex
+++ b/lib/xhttp/headers.ex
@@ -1,0 +1,121 @@
+defmodule XHTTP.Headers do
+  @moduledoc ~S"""
+  Functions for working with HTTP request and response headers, as described
+  in the [HTTP 1.1 specification](https://www.w3.org/Protocols/rfc2616/rfc2616.html).
+
+  Headers are represented in Elixir as a list of `{"header_name", "value"}`
+  tuples.  Multiple entries for the same header name are allowed.
+
+  Capitalization of header names is preserved during insertion
+  (`put_header/3`), however header names are handled case-insensitively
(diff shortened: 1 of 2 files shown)


whatyouhide:
Hey @gamache! Thanks a lot for the PR :D

This library is at its very early stages and we're moving very fast (and breaking a lot), so we're not ready to think about public APIs yet. We'll have to have a more complete picture of the implementation before being able to get a good idea on how to abstract functionalities.

Please do bring up this discussion later on in the development though :) Thanks! 💟

gamache:
Thanks for the comments @whatyouhide!  I know what you mean about moving fast and breaking a lot; I've been writing a [high-level HTTP client called XClient](https://github.com/appcues/xclient) on top of yours.  :smiling_imp:  (Bet you didn't think you'd have downstream projects before you published to Hex, did you?)

Anyway, I've stashed this stuff in `XClient.Headers` for now.  I'm happy to poke my head back in here when it's time to discuss a public API.

PR #28, opened 2018-04-09
Make transport operations stateful

whatyouhide:
I left some operations (like `getopts` and `setopts`) stateless for now as I don't think we'll need to make them stateful. Also, I left TCP/SSL messages untouched for now as I think we'll need to add callbacks for streaming messages (so that for example a proxy conn transport can delegate the streaming to the underlying TCP/SSL transport).

\cc @josevalim as always <3

### DIFF
--- a/lib/xhttp/transport.ex
+++ b/lib/xhttp/transport.ex
@@ -6,8 +6,7 @@ defmodule XHTTP.Transport do
   @callback connect(host :: String.t(), port :: :inet.port_number(), opts :: keyword()) ::
               {:ok, state()} | error()
 
-  @callback negotiated_protocol(state()) ::
-              {:ok, protocol :: binary()} | {:error, reason :: term()}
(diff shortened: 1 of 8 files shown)


lib/xhttp/transport.ex
```diff
@@ -1,23 +1,23 @@
...
 
-  @callback connect(host :: String.t(), port :: :inet.port_number(), opts :: Keyword.t()) ::
-              {:ok, state()} | {:error, reason :: term()}
+  @type error() :: {:error, reason :: term()}
+
+  @callback connect(host :: String.t(), port :: :inet.port_number(), opts :: keyword()) ::
+              {:ok, state()} | error()
 
   @callback negotiated_protocol(state()) ::
               {:ok, protocol :: binary()} | {:error, reason :: term()}
```
michalmuskala: This should probably use `error()` for consistency.

whatyouhide: Good catch! Updated.

PR #59, opened 2019-01-30
Refactor the docs and move most of them to XHTTP

whatyouhide:
We're now missing an "Architecture" page in the docs as far as documentation goes.

### DIFF
--- a/lib/xhttp.ex
+++ b/lib/xhttp.ex
@@ -157,11 +157,13 @@ defmodule XHTTP do
   Once a proxied connection is returned, the proxy is transparent to you and you
   can use the connection like a normal HTTP/1.1 connection.
 
-  If the `scheme` is `:http`, we will use an unsafe proxy to connect to
-  the given host.
+  If the `scheme` is `:http`, we will connect to the host in the most compatible
+  way, supporting older proxy servers. Data will be sent in clear text.
 
-  If the scheme is `:https`, we will use a tunnel proxy to connect to the
-  given host.
+  If the connection scheme is `:https`, we will connect to the host with a tunnel
+  through the proxy. Using `:https` for both the proxy and the connection scheme
+  is not supported, it is recommended to use `:https` for the end host connection
+  instead of the proxy.
 
   ## Examples
 
--- a/lib/xhttp.ex
+++ b/lib/xhttp.ex
@@ -72,6 +72,7 @@ defmodule XHTTP do
 
   As you can see, all responses have the unique request reference as the second
   element of the tuple, so that we know which request the response belongs to.
+  See `t:XHTTP.responses/0` for the full list of responses returned by `XHTTP.stream/2`.
 
   ## Architecture
 
--- a/lib/xhttp.ex
+++ b/lib/xhttp.ex
@@ -79,7 +79,7 @@ defmodule XHTTP do
   we use this HTTP client. Usually, you will want to create this data structure
   in a process that acts as *connection manager*. Sometimes, you might want to
   have a single process responsible for multiple connections, either to just one
-  host or different hosts. For more discussion on architectures based off of this
+  host or multiple hosts. For more discussion on architectures based off of this
   HTTP client, see the [TODO architecture] page in the docs.
   """
 
@@ -105,7 +105,7 @@ defmodule XHTTP do
   identified by the given `host` and `port` combination. Both HTTP and HTTPS are supported
   by passing respectively `:http` and `:https` as the `scheme`.
 
-  The connection struct wraps a TCP/SSL socket, which is created once the connection
+  The connection struct wraps a socket, which is created once the connection
   is established inside this function. If HTTP is used, then the created socket is a TCP
   socket and the `:gen_tcp` module is used to create that socket. If HTTPS is used, then
   the created socket is an SSL socket and the `:ssl` module is used to create that socket.
@@ -168,7 +168,7 @@ defmodule XHTTP do
 
   Using a proxy:
 
-      proxy = {:https, "myproxy.example.com", 443, []}
+      proxy = {:http, "myproxy.example.com", 80, []}
       {:ok, conn} = XHTTP.connect(:https, "httpbin.org", 443, proxy: proxy)
 
   Forcing the connection to be an HTTP/2 connection:
--- a/lib/xhttp.ex
+++ b/lib/xhttp.ex
@@ -1,7 +1,86 @@
 defmodule XHTTP do
   @moduledoc """
-  Single interface for `XHTTP1` and `XHTTP2` with version negotiation support and support for
-  proxies.
+  Processless HTTP connection data structure and functions.
+
+  Single interface for `XHTTP1` and `XHTTP2` with version negotiation support
+  and support for proxies.
+
+  ## Usage
+
+  To establish a connection with a given server, use `connect/4`. This will
+  return an opaque data structure, `%XHTTP{}`, that represents the connection
+  to the server. To send a request, you can use `request/5`. Sending a request
+  does not take care of the response to that request. The connection is a
+  wrapper around a TCP (`:gen_tcp` module) or SSL (`:ssl` module) socket that is
+  set in **active mode**. This means that TCP/SSL messages will be delivered to
+  the process that started the connection and created the `%XHTTP{}` data structure.
+
+  The process that owns the connection is responsible for receiving the messages
+  (for example, a GenServer is responsible for defining `handle_info/2`). However,
+  `XHTTP` makes it easy to identify TCP/SSL messages that are coming from the
+  connection with the server with the `stream/2` function. This function takes the
+  connection and a term and returns `:unknown` if the term is not a TCP/SSL message
+  belonging to the connection. If the term *is* a message for the connection, then
+  a response and a new connection are returned. It's important to store the new
+  returned connection data structure over the old one since the connection is an
+  immutable data structure.
+
+  Let's see an example of a common workflow of connecting to a server, sending a
+  request, and processing the response. We start by using `connect/3` to connect
+  to a server.
+
+      {:ok, conn} = XHTTP.connect(:http, "httpbin.org", 80)
+
+  `conn` is a `%XHTTP{}` data structure that represents the connection. This
+  data structure is a wrapper around a TCP (`:gen_tcp` module) or SSL (`:ssl`
+  module) socket that is set in **active mode**. This means that TCP/SSL messages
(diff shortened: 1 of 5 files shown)


lib/xhttp.ex
```diff
@@ -19,6 +98,84 @@ defmodule XHTTP do
...
+  You can set up proxying through the `:proxy` option, which is a tuple
+  `{scheme, hostname, port, opts}` that identifies the proxy to connect to.
+  Once a proxied connection is returned, the proxy is transparent to you and you
+  can use the connection like a normal HTTP/1.1 connection.
+
+  If the `scheme` is `:http`, we will use an unsafe proxy to connect to
+  the given host.
+
+  If the scheme is `:https`, we will use a tunnel proxy to connect to the
+  given host.
```
ericmj: If the connection scheme is `:https`, we will connect to the host with a tunnel through the proxy.  Using `:https` for both the proxy and the connection scheme is not supported, it is recommended to use `:https` for the end host connection instead of the proxy.

lib/xhttp.ex
```diff
@@ -19,6 +98,84 @@ defmodule XHTTP do
...
+
+  ## Examples
+
+      {:ok, conn} = XHTTP.connect(:http, "httpbin.org", 80)
+
+  Using a proxy:
+
+      proxy = {:https, "myproxy.example.com", 443, []}
```
ericmj:
```suggestion
      proxy = {:http, "myproxy.example.com", 80, []}
```

lib/xhttp.ex
```diff
@@ -69,10 +231,74 @@ defmodule XHTTP do
...
     do: XHTTP.Negotiate.initiate(transport, transport_state, hostname, port, opts)
 
+  @doc """
+  Checks whether the connection is open.
+
+  This function returns `true` if the connection is open, `false` otherwise. It should
+  be used to check that a connection is open before sending requests or performing
+  operations that involve talking to the server.
+
+  If a connection is not open, it has become useless and you should get rid of it.
```
ericmj: :)

lib/xhttp.ex
```diff
@@ -69,10 +231,74 @@ defmodule XHTTP do
...
+  In HTTP/1.1, you can't open a request if the body of another request is streaming.
+  See `XHTTP1.request/5` for more information.
+
+  For a quick discussion on HTTP/2 streams and requests, see the `XHTTP2` module and
+  `XHTTP2.request/5`.
+
+  ## Examples
+
+      XHTTP.request(conn, "GET", "/", _headers = [])
+      XHTTP.request(conn, "POST", "/path", [{"Content-Type", "application/json"}], "{}")
```
ericmj:
```suggestion
      XHTTP.request(conn, "POST", "/path", [{"content-type", "application/json"}], "{}")
```

I was thinking we should use lower case headers to promote that, not super important...

lib/xhttp.ex
```diff
@@ -86,32 +312,223 @@ defmodule XHTTP do
...
+      request. Usually the server won't send any reply until this is sent.
+
+  This function always returns an updated connection to be stored over the old connection.
+
+  ## Examples
+
+  Let's see an example of streaming an empty JSON object (`{}`) by streaming one curly
+  brace at a time.
+
+      headers = [{"Content-Type", "application/json"}, {"Content-Length", "2"}]
```
ericmj:
```suggestion
      headers = [{"content-type", "application/json"}, {"content-length", "2"}]
```

Lower case headers again.

lib/xhttp.ex
```diff
@@ -86,32 +312,223 @@ defmodule XHTTP do
...
+  `request_ref` returned by `request/5` to identify the request to stream the body for,
+  and a chunk of body to stream. The value of chunk can be:
+
+    * iodata - a chunk of iodata is transmitted to the server as part of the body
+      of the request.
+
+    * `:eof` - signals the end of the streaming of the request body for the given
+      request. Usually the server won't send any reply until this is sent.
+
+  This function always returns an updated connection to be stored over the old connection.
```
ericmj: Lets add a paragraph:

When streaming the request body xhttp can not send a precalculated `content-length` request header. It is up to you set the correct headers depending on how you stream the body, either by setting the `content-length` header yourself or by using the appropriate transfer encoding if using HTTP/1.

(We don't have transfer encoding with http/2, right?)

lib/xhttp.ex
```diff
@@ -86,32 +312,223 @@ defmodule XHTTP do
...
   @impl true
   @spec stream_request_body(t(), request_ref(), iodata() | :eof) ::
           {:ok, t()} | {:error, t(), term()}
   def stream_request_body(conn, ref, body),
     do: conn_module(conn).stream_request_body(conn, ref, body)
 
+  @doc """
+  Streams the next batch of responses from the given message.
+
+  This function processes a "message" which can be any term, but should be
```
ericmj: Not any term, right? We document it is as the `socket_message` type/

lib/xhttp.ex
```diff
@@ -86,32 +312,223 @@ defmodule XHTTP do
...
+  response-specific terms as well, documented below.
+
+  These are the possible responses that can be returned.
+
+    * `{:status, request_ref, status_code}` - returned when the server replied
+      with a response status code. The status code is a non-negative integer.
+
+    * `{:headers, request_ref, headers}` - returned when the server replied
+      with a list of headers. Headers are in the form `{header_name, header_value}`
+      with `header_name` and `header_value` being strings.
```
ericmj: Add a note that status and headers will be returned once and only once per request?

lib/xhttp.ex
whatyouhide: Definitely agree on that.

lib/xhttp.ex
whatyouhide: Done!

lib/xhttp.ex
```diff
@@ -1,7 +1,86 @@
...
+
+  `responses` is a list of possible responses. The most common responses are:
+
+    * `{:status, request_ref, status_code}` for the status code
+    * `{:headers, request_ref, headers}` for the response headers
+    * `{:data, request_ref, binary}` for pieces of the response body
+    * `{:done, request_ref}` for the end of the response
+
+  As you can see, all responses have the unique request reference as the second
+  element of the tuple, so that we know which request the response belongs to.
```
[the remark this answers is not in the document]
whatyouhide: Done as a commit to split this into multiple lines.

lib/xhttp.ex
```diff
@@ -19,6 +98,84 @@ defmodule XHTTP do
...
+      list, then XHTTP will negotiate the protocol. See the section "Protocol negotiation"
+      below for more information. Defaults to `[:http1, :http2]`.
+
+  The following options are HTTP/1.1-specific and will force the connection
+  to be an HTTP/1.1 connection.
+
+    * `:proxy` - a `{scheme, hostname, port, opts}` tuple that identifies a proxy to
+      connect to. See the "Proxying" section below for more information.
+
+  The following options are HTTP/2-specific and will only be used on HTTP/2 connections.
```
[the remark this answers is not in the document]
whatyouhide: I'm using "connection**s**" so I think we're fine without the "a". :)

lib/xhttp.ex
whatyouhide: We don't. Added this paragraph.

lib/xhttp.ex
whatyouhide: Hrm, you're right that we document it like that but it's wrong since we return `:unknown` for any message that is not a socket message. Would this spec work?

```elixir
@spec stream(t(), socket_message()) :: {:ok, ...} | {:error, ...}
@spec stream(t(), term()) :: :unknown
```

? It might not since `term/0` is a superset of `socket_message/0`.

lib/xhttp.ex
```diff
@@ -19,6 +98,84 @@ defmodule XHTTP do
...
+
+  ## Proxying
+
+  You can set up proxying through the `:proxy` option, which is a tuple
+  `{scheme, hostname, port, opts}` that identifies the proxy to connect to.
+  Once a proxied connection is returned, the proxy is transparent to you and you
+  can use the connection like a normal HTTP/1.1 connection.
+
+  If the `scheme` is `:http`, we will use an unsafe proxy to connect to
+  the given host.
```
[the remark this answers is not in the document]
whatyouhide: Addressed.

lib/xhttp.ex
whatyouhide: Addressed.

lib/xhttp.ex
```diff
@@ -19,6 +98,84 @@ defmodule XHTTP do
...
+  identified by the given `host` and `port` combination. Both HTTP and HTTPS are supported
+  by passing respectively `:http` and `:https` as the `scheme`.
+
+  The connection struct wraps a TCP/SSL socket, which is created once the connection
+  is established inside this function. If HTTP is used, then the created socket is a TCP
+  socket and the `:gen_tcp` module is used to create that socket. If HTTPS is used, then
+  the created socket is an SSL socket and the `:ssl` module is used to create that socket.
+  The socket is created in active mode, which is why it is important to know the type of
+  the socket: messages from the socket will be delivered directly to the process that
+  creates the connection and tagged appropriately (see the `:gen_tcp` and `:ssl` modules).
```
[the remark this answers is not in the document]
whatyouhide: Addressed.

lib/xhttp.ex
```diff
@@ -19,6 +98,84 @@ defmodule XHTTP do
   @type socket_message() :: XHTTPCore.Conn.socket_message()
```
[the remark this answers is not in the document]
whatyouhide: They are not linked by ExDoc (since the module they should link to is not in the docs).

image

I would suggest to take care of typespecs in a separate PR (where I'd like to document the types as well).

lib/xhttp.ex
```diff
@@ -1,7 +1,86 @@
...
+  Let's see an example of a common workflow of connecting to a server, sending a
+  request, and processing the response. We start by using `connect/3` to connect
+  to a server.
+
+      {:ok, conn} = XHTTP.connect(:http, "httpbin.org", 80)
+
+  `conn` is a `%XHTTP{}` data structure that represents the connection. This
+  data structure is a wrapper around a TCP (`:gen_tcp` module) or SSL (`:ssl`
+  module) socket that is set in **active mode**. This means that TCP/SSL messages
+  will be delivered to the process that started the connection.
```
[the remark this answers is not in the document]
whatyouhide: It was not on purpose. Addressed :)

PR #70, opened 2019-02-12
Add the start of an "Architecture" page in the docs

whatyouhide:
This will not show up in the docs until #65 is merged, otherwise we would have conflicts in the docs configuration in the `mix.exs` file.


pages/Architecture.md
```diff
@@ -0,0 +1,105 @@
...
+      {:ok, state}
+    end
+  end
+
+  @impl true
+  def handle_call({:request, method, path, headers, body}, from, state) do
+    # In both the successful case and the error case, we make sure to update the connection
+    # struct in the state since the connection is an immutable data structure.
+    case Mint.HTTP.request(state.conn, method, path, headers, body) do
+      {:ok, request_ref, conn} ->
```
fertapric: I think this should be `{:ok, conn, request_ref}` and `{:error, conn, reason}`.

https://github.com/ericmj/mint/blob/fb749eefa47e21821a06d7e013ca4e73684cdca8/lib/mint/http.ex#L325-L327

PR #107, opened 2019-03-05
Improve HTTP/2 test server

whatyouhide:
Sometimes we would be okay with frames coming to the server after the test was done. This way we wait for the server to confirm that no data is left and no frames are left to handle.


whatyouhide:
Hopefully this fixes the build! 😄

PR #120, opened 2019-03-08
Improve storing of HTTP/2 client vs server settings

### LINKED ISSUE #118: Differentiate between retrieving client vs server settings in Mint.HTTP2.get_setting/2
Right now we retrieve the server settings but there's no way to retrieve the client settings.

whatyouhide:
Closes #118.

Before this change, we stored the client settings and the server settings at the same level in the connection, with a weird naming scheme and generally a pretty confusing approach. This change introduces the keys `:client_settings` and `:server_settings` to the `Mint.HTTP2` connection struct so that it's always crystal-clear which one we're referring to.

### DIFF
--- a/lib/mint/http2.ex
+++ b/lib/mint/http2.ex
@@ -25,6 +25,32 @@ defmodule Mint.HTTP2 do
   This is why we identify each request with a unique reference returned by `request/5`.
   See `request/5` for more information.
 
+  ## HTTP/2 settings
+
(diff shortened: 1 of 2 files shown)


whatyouhide:
I actually improved on this and implemented fixes for #118 as well. I removed `Mint.HTTP2.get_setting/2` (which retrieved server settings) in favour of `Mint.HTTP2.get_server_setting/2` and `Mint.HTTP2.get_client_setting/2`.

PR #137, opened 2019-03-18
Simplify HTTP/1 header parsing and test it through property-based testing

### LINKED ISSUE #126: Use property-based testing on HTTP/1 header formatting

whatyouhide:
Closes #126.

I also removed cross-module throw/catch instances coming from the `Mint.HTTP1.Parse` module, similar to what was done in #136.

### DIFF
--- a/lib/mint/http1/parse.ex
+++ b/lib/mint/http1/parse.ex
@@ -13,32 +13,7 @@ defmodule Mint.HTTP1.Parse do
     end
   end
 
-  defp lower_char(?A), do: ?a
-  defp lower_char(?B), do: ?b
-  defp lower_char(?C), do: ?c
-  defp lower_char(?D), do: ?d
-  defp lower_char(?E), do: ?e
-  defp lower_char(?F), do: ?f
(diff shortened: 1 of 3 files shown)


lib/mint/http1/parse.ex
```diff
@@ -13,32 +13,7 @@ defmodule Mint.HTTP1.Parse do
...
-  defp lower_char(?R), do: ?r
-  defp lower_char(?S), do: ?s
-  defp lower_char(?T), do: ?t
-  defp lower_char(?U), do: ?u
-  defp lower_char(?V), do: ?v
-  defp lower_char(?W), do: ?w
-  defp lower_char(?X), do: ?x
-  defp lower_char(?Y), do: ?y
-  defp lower_char(?Z), do: ?z
+  defp lower_char(char) when char in ?A..?Z, do: char + 32
```
whatyouhide: @ericmj not sure you where doing this to optimize the pattern matching but we're losing quite a lot of lines of code so we might be better off with this anyways.

PR #140, opened 2019-03-18
Simplify and improve errors in Mint.HTTP1.Parse

### LINKED ISSUE #138: Improve errors returned by HTTP/1 header parsing

whatyouhide:
Closes #138.

I removed all throw/catches in `Mint.HTTP1.Parse` as they were unnecessary. I also improved the `:invalid_token_list` error when parsing HTTP/1 header values so that it shows the offending string. I finally added formatting to the `:invalid_token_list` and `:empty_token_list` errors in `Mint.HTTP1`.

### DIFF
--- a/lib/mint/http1.ex
+++ b/lib/mint/http1.ex
@@ -575,8 +575,7 @@ defmodule Mint.HTTP1 do
     %{conn | state: :closed}
   end
 
-  # TODO: We should probably error if both transfer-encoding and content-length
-  # is set. RFC7230 3.3.3:
+  # RFC7230 3.3.3:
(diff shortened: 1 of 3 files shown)


lib/mint/http1.ex
```diff
@@ -575,8 +575,7 @@ defmodule Mint.HTTP1 do
     %{conn | state: :closed}
   end
 
-  # TODO: We should probably error if both transfer-encoding and content-length
```
whatyouhide: @ericmj we do return an error here, so we can remove this TODO, right?

PR #146, opened 2019-03-20
Add specs to the Mint.HTTPError exception

whatyouhide:


whatyouhide:
@ericmj I'd rather keep them in the moduledoc. Exceptions are already slightly harder to find in the documentation since they are not under "Modules" (they're under "Exceptions"), so I think it's better to avoid adding one more level of documentation to find and leave them in the module documentation.

PR #149, opened 2019-03-23
Fix a bug in Mint.TunnelProxy

whatyouhide:
We were not handling the `{:error, conn, reason}` case but at the same time we were not using `conn` in `handle_responses/4`, so I removed the passing around of the connection to make things simpler.

### DIFF
--- a/lib/mint/tunnel_proxy.ex
+++ b/lib/mint/tunnel_proxy.ex
@@ -63,7 +63,7 @@ defmodule Mint.TunnelProxy do
   defp stream(conn, ref, timeout_deadline, msg) do
     case HTTP1.stream(conn, msg) do
       {:ok, conn, responses} ->
-        case handle_responses(conn, ref, timeout_deadline, responses) do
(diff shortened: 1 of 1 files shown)


lib/mint/tunnel_proxy.ex
```diff
@@ -63,7 +63,7 @@ defmodule Mint.TunnelProxy do
   defp stream(conn, ref, timeout_deadline, msg) do
     case HTTP1.stream(conn, msg) do
       {:ok, conn, responses} ->
-        case handle_responses(conn, ref, timeout_deadline, responses) do
+        case handle_responses(ref, timeout_deadline, responses) do
           :done -> :ok
           :more -> receive_response(conn, ref, timeout_deadline)
           {:error, reason} -> {:error, conn, reason}
```
whatyouhide: @ericmj the bug was here. We were returning `{:error, conn, reason}` from `handle_responses/4` but not handling it here. Now we return `{:error, reason}` so this is fine.

PR #161, opened 2019-04-11
Fix bug with HTTP/2 requests exceeding window size

whatyouhide:
Before this commit, we had a bug when sending HTTP/2 requests with a body exceeding the max window size of either the stream or the connection. Basically, we were sending a HEADERS frame and only then checking for the window size and possibly error out. This meant that sending a request that exceeded the window size resulted in sending the HEADERS and then erroring out, leaving the request in an open state on the server.

With this commit, we build the payload for HEADERS and DATA without sending it right away, and send them together if everything is correct (which also leads to an optimization since we do only one socket send call instead of two when sending requests with headers and body).

I also fixed a small bug when checking the window size: before we were checking if the size of the request body was greater or equal to the window sizes, but the right check is just strictly greater than the window size. Now, if you have a window size of 5 and send a body of 5 bytes, it will work correctly.

@ericmj I am experiencing this log coming from SSL from time to time:

```
18:03:30.723 [info]  ['TLS', 32, 'client', 58, 32, 73, 110, 32, 115, 116, 97, 116, 101, 32, 'connection', 32, 'received SERVER ALERT: Fatal - Internal Error', 10]
```

I am unsure how these changes could have caused this and no idea what it is. Thoughts?

**Edit**: https://travis-ci.org/ericmj/mint/jobs/518843663#L662

### DIFF
--- a/lib/mint/http2.ex
+++ b/lib/mint/http2.ex
@@ -479,15 +479,17 @@ defmodule Mint.HTTP2 do
     conn =
       case body do
         :stream ->
-          send_headers(conn, stream_id, headers, [:end_headers])
+          {conn, payload} = encode_headers(conn, stream_id, headers, [:end_headers])
+          send!(conn, payload)
 
         nil ->
-          send_headers(conn, stream_id, headers, [:end_stream, :end_headers])
+          {conn, payload} = encode_headers(conn, stream_id, headers, [:end_stream, :end_headers])
+          send!(conn, payload)
 
         _iodata ->
-          # TODO: Optimize here by sending a single packet on the network.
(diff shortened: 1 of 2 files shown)


whatyouhide:
Actually maybe pinging @voltone might solve the TLS issue 😄

voltone:
I can reproduce it with OTP 21.3, and when I add some `:timer.sleep/1` at the end of the test case it becomes consistent. With OTP 21.3.2 I don't see it.

It is likely due to OTP-14701, which is a server-side bug that was fixed in 21.3.1. Since it only affects the TestServer and not Mint itself I wouldn't worry about.

whatyouhide:
Awesome, thanks so much @voltone! 💟

PR #179, opened 2019-06-18
Support trailing response headers in HTTP/2

whatyouhide:
This change required to reword the way we handle incoming `HEADERS` (plus possible `CONTINUATION`) frames in HTTP/2. Before this commit, we would decode the headers and then for example check that the :status header was present.

Now, we need to track whether the first bunch of headers (the *request* headers) were received for each stream. If they were received, then the next headers that come for that stream have to be trailing headers and have to have the `END_STREAM` flag set (otherwise we do a protocol error).

### DIFF
--- a/lib/mint/http2.ex
+++ b/lib/mint/http2.ex
@@ -1070,7 +1070,8 @@ defmodule Mint.HTTP2 do
       id: conn.next_stream_id,
       ref: make_ref(),
       state: :idle,
-      window_size: conn.server_settings.initial_window_size
+      window_size: conn.server_settings.initial_window_size,
+      received_first_headers?: false
     }
 
     conn = put_in(conn.streams[stream.id], stream)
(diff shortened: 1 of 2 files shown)


lib/mint/http2.ex
```diff
@@ -1070,7 +1070,8 @@ defmodule Mint.HTTP2 do
       id: conn.next_stream_id,
       ref: make_ref(),
       state: :idle,
-      window_size: conn.server_settings.initial_window_size
+      window_size: conn.server_settings.initial_window_size,
+      received_first_headers?: false
```
whatyouhide: We need to track whether the first "batch" of headers (the **response headers**) were received. Couldn't think of a better way.

lib/mint/http2.ex
```diff
@@ -1466,16 +1467,29 @@ defmodule Mint.HTTP2 do
...
+      _ = Logger.debug(fn -> "Received HEADERS frame on closed stream ID" end)
+      {conn, responses}
+    end
+  end
+
+  defp handle_decoded_headers_for_stream(conn, responses, stream, headers, end_stream?) do
+    %{ref: ref, received_first_headers?: received_first_headers?} = stream
+
+    case headers do
+      [{":status", status} | headers] when not received_first_headers? ->
```
whatyouhide: This is why we track whether headers were received. If we didn't receive headers yet (so these are the **response headers**), we need to check whether the `:status` header is there. For trailing headers, we don't.

PR #181, opened 2019-06-21
Prevented unnecessary buffer allocation

scohen:
If the existing body buffer is empty, we can safely set the incoming body to
be the initial state of that buffer. This prevents a rather slow
allocation from taking place.
In my testing in a database driver, this speeds up the query's
response processing by over two seconds.

### DIFF
--- a/lib/mint/core/util.ex
+++ b/lib/mint/core/util.ex
@@ -41,4 +41,10 @@ defmodule Mint.Core.Util do
 
   def downcase_ascii_char(char) when char in ?A..?Z, do: char + 32
   def downcase_ascii_char(char) when char in 0..127, do: char
+
+  # If the buffer is empty, reusing the incoming data saves
+  # a potentially large allocation of memory.
+  # This should be fixed in a subsequent OTP release.
+  def maybe_concat(<<>>, data), do: data
+  def maybe_concat(buffer, data) when is_binary(buffer), do: buffer <> data
 end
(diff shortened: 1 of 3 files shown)


michalmuskala:
I think OTP 22 shouldn't reallocate in a case like this.

whatyouhide:
@scohen great catch. Can you:

1. Make sure that this doesn't happen in HTTP/2?
2. Leave a comment that that should be removed when we'll support OTP 22+ as @michalmuskala mentioned?

scohen:
@michalmuskala I'm presently running otp 22:
```
Erlang/OTP 22 [erts-10.4.1] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe] [dtrace]

Interactive Elixir (1.8.2) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)>
```
And I will make the requisite changes in the http2 module as well.

michalmuskala:
I'm sorry, my comment actually wasn't true. OTP 22 will optimize appends of an empty strings, but not prepends. This is unfortunately related to how the bytecodes for binary concatenation look like and the prepend case is hard to optimise.

For reference this is the PR that introduced this optimisation https://github.com/erlang/otp/pull/2055

whatyouhide:
Beautiful, thanks a lot @scohen! 💟

PR #182, opened 2019-06-23
Remove double wrapping of transport error by receive

liveforeverx:
Any direct operation comes already with wrapped error,
there is no need to wrap it again.


whatyouhide:
Thanks @liveforeverx! 💟

PR #186, opened 2019-07-10
Add support for trailing headers in HTTP requests

### LINKED ISSUE #180: Support trailing headers in requests
## The problem

Right now (especially after #179), we support trailing headers in responses (both HTTP/1 and HTTP/2). They're returned as a normal `:headers` response after the `:data` (and there can only be one of them). However, we don't support *sending* trailing headers in requests.

## Solution

@ericmj discussed a solution. Right now, we can stream the request body with `stream_request_body/3`. When you're done streaming the body, you can pass `:eof` to the `stream_request_body/3` function as the body. Our solution consists in adding a `{:eof, trailing_headers}` possible body value. For example:

```elixir
{:ok, conn, ref} = # open request
stream_request_body(conn, ref, "foo")
stream_request_body(conn, ref, {:eof, [{"x-trailing", "some value"}]})
```

For now, we thought about keeping the `stream_request_body` name because trailing headers are kind of part of the request body and so that we don't need to change the name for a smaller thing. However, we can consider changing the name of this function to something else if the community prefers that way.

whatyouhide:
Closes #180.

### DIFF
--- a/lib/mint/core/conn.ex
+++ b/lib/mint/core/conn.ex
@@ -30,7 +30,7 @@ defmodule Mint.Core.Conn do
   @callback stream_request_body(
               conn(),
               Types.request_ref(),
-              body_chunk :: iodata() | :eof | {:eof, Types.headers()}
+              body_chunk :: iodata() | :eof | {:eof, trailing_headers :: Types.headers()}
             ) ::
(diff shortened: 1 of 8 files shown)


whatyouhide:
This is ready to review @ericmj, added docs and everything :).

PR #196, opened 2019-07-14
Update Architecture.md

jmerriweather:
The example wouldn't work for me as the variable `data` was overwritten.

Changed the two `data` variables into `existing_data` and `new_data`


whatyouhide:
Great catch @jmerriweather, thank you! 💟

PR #201, opened 2019-08-04
Fix architecture example

jayjun:
GenServer’s `init` callback expects `{:stop, reason}` during failures.

Current example returns `{:error, reason}` which logs an inexact error.

```
** (EXIT from #PID<0.182.0>) shell process exited with reason: bad return value: {:error, %Mint.TransportError{reason: :timeout}}
```


whatyouhide:
Thanks @jayjun! 💟

PR #202, opened 2019-08-08
Issue #200: Don't include default port in :authority

jmmk:


whatyouhide:
Thanks a lot @jmmk! 💟 I'm merging this and then improving the tests slightly so that we avoid changing the global default port.

PR #254, opened 2020-04-25
Remove mention of `t:socket_message/0`

wojtekmach:
The type is long gone!


whatyouhide:
Nice catch, thanks @wojtekmach! 💟

PR #277, opened 2020-09-17
Fix function names for decompression guide

ahamez:
Function `decompress_data/2` delegates the real work via `Enum.reduce` to `decompress_with_algorithm/2`.


whatyouhide:
Great catch, thanks @ahamez! 💟

PR #280, opened 2020-09-29
Fix connection closed errors not propagating, closes #279

### LINKED ISSUE #279: Connection errors not being surfaced from `stream`.
Hey y'all,

So I've been trying to track down the some of the issues with with mint returning socket closed errors as discussed here: https://github.com/keathley/finch/issues/62. As part of that it seems like Mint error when setting the socket back to `active: :once` after streaming, but not return that result to the caller: https://github.com/elixir-mint/mint/blob/master/lib/mint/http2.ex#L837-L838. In my tests we can get a socket closed error here, but since the conn isn't being updated we have no way to detect that. This leaves Finch's connection state machine open, despite the underlying connection having been closed. Since the connection is closed, we receive no further messages and we don't detect the error until the caller attempts to re-use that connection, at which point they receive an error.

I'm happy to make the code change here, but I wanted to see what y'all thought we should be returning here. It looks like this behaviour is the same for both http1 and http2, so both will need to be updated.

josevalim:
There is one failing test but I believe the test is wrong. We are transferring the test to another process and then trying to stream its response, which fails when trying to set it back to active mode. I want to change the test but I would like confirmation before.

### DIFF
--- a/lib/mint/http1.ex
+++ b/lib/mint/http1.ex
@@ -265,7 +265,7 @@ defmodule Mint.HTTP1 do
   @doc """
   See `Mint.HTTP.stream_request_body/3`.
 
-  In HTTP/1, sending an empty chuunk is a no-op.
+  In HTTP/1, sending an empty chunk is a no-op.
 
   ## Transfer encoding and content length
 
--- a/lib/mint/http1.ex
+++ b/lib/mint/http1.ex
@@ -384,14 +384,17 @@ defmodule Mint.HTTP1 do
 
   def stream(%__MODULE__{transport: transport, socket: socket} = conn, {tag, socket, data})
       when tag in [:tcp, :ssl] do
-    result = handle_data(conn, data)
+    case handle_data(conn, data) do
+      {:ok, %{mode: mode, state: state} = conn, responses}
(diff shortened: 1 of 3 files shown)


whatyouhide (approved):
This looks good and the test seems indeed wrong, I'm okay with changing it :)x

josevalim:
@ericmj helped me fixed the test. I have also added a test for the functionality that we need.

sneako:
FWIW: I deployed this change to an application which uses Finch and makes a ton of http1 requests (no http2). I am not observing any new errors or regressions and the rate of connection closed errors that we have been trying to track down [here](https://github.com/keathley/finch/issues/62) is practically the same as before.

josevalim:
@sneako are you sure the server does not support http2? Because due to https://github.com/keathley/finch/pull/87, we may be accidentally using http2 and that could have been the root cause. :)

sneako:
@josevalim We have a dozen pools, and one pool in particular emits almost all of the connection closed errors, however I just took a packet capture to confirm that we are in fact using HTTP/1.1 for this host.

keathley:
Tested locally, and this PR resolves the issues in the Finch tests that I was seeing.

PR #283, opened 2020-10-27
Support all types of addresses (including unix domain sockets aka local addresses)

myobie:
Supporting all types of addresses means accepting `inet:socket_address() | inet:hostname()` types like [`gen_tcp:connect/3`](http://erlang.org/doc/man/gen_tcp.html#connect-3).

## TODO

- [x] Rename hostname parameter to address and accept all address types the underlying transports accept.
- [x] Add a :hostname connection option. This option will be used for the Host header, and passed down as an option to the transports to use for hostname verification, SNI, etc.
- [x] The :hostname option is required if address is not a string, otherwise it defaults to the same value as address.
- [x] Make sure all docs are changed to read `address` instead of `hostname`

Examples:

```ex
# unix domain socket
{:ok, conn} = Mint.HTTP.connect(:http, {:local, "/tmp/consul.sock"}, 0, hostname: "localhost")

# loopback
{:ok, conn} = Mint.HTTP.connect(:http, :loopback, 80, hostname: "localhost")

# ip address
{:ok, conn} = Mint.HTTP.connect(:http, {93, 184, 216, 34}, 80, hostname: "example.com")
```

---

<details>
<summary>Original PR description about only unix domain sockets</summary>

This PR adds support for unix domain sockets by accepting a special hostname prefixed with `unix://` to indicate it's an [`inet:local_address`](http://erlang.org/doc/man/inet.html#type-local_address).

_Aside: I believe Mint should eventually accept all types of addresses supported by [`gen_tcp:connect/3`](http://erlang.org/doc/man/gen_tcp.html#connect-3); however, that would be a much larger refactor which more-than-likely renames `hostname` to `address` and adds references to the address types from `inet`._

Example usage could be:

```ex
{:ok, conn} = Mint.HTTP.connect(:http, "unix:///tmp/consul.sock", 0)
```

While I have tested this "live," I am unsure how to test this programmatically and would welcome any help getting that setup. I did notice [the PR which added ipv6 support](https://github.com/elixir-mint/mint/pull/270/files) didn't add additional tests so maybe it's alright to go without them?

</details>


ericmj:
Does cowboy support connecting over unix sockets? If so you can add a test the connects to a cowboy server.

Won't the `Host` header be incorrect unless the user manually sets it? Does `:ssl` support unix sockets?

myobie:
@ericmj good questions.

Yes it will be incorrect if it's not set by the user and that is expected. Not every server is going to validate the host header. `curl` has a advantage for this since one must specify a complete URL so it can take it out of that URL (both of the below approaches work):

```sh
curl --unix-socket /tmp/consul.sock http://localhost/v1/agent/members
# or
curl --unix-socket /tmp/consul.sock -H'Host: localhost' http://-/v1/agent/members
```

It seems [cowboy should support unix domain sockets](https://github.com/ninenines/cowboy/issues/803) so I will try writing a test case for that. About `:ssl`, I _assume_ it works no matter the type of socket but it is something a test case would help make for certain.

ericmj:
We have discussed options for passing an overriding address in this issue: https://github.com/elixir-mint/mint/pull/282#issuecomment-703535702. We could potentially pass the unix socket name the same way.

Another option is to do the refactor from hostname to address like you are suggesting. With that refactor we should also add a `:hostname` option that will be used for the `Host` header and for SSL hostname verification and SNI.

/cc @whatyouhide

myobie:
Note: `{:local, path}` doesn't seem to be valid for OTP <19: https://github.com/elixir-mint/mint/pull/283/checks?check_run_id=1317106396#step:6:601

myobie:
@ericmj I've pushed a test file which tests both transports (so `:gen_tcp` and `:ssl`) and they both work with unix sockets. I'll figure out how to tell the older OTP CI to skip the tests, but is there anything else you think is missing from this? Thanks 🙏

ericmj:
I don't think we should add a `:hostname` transport option that only the SSL transport uses and does not affect the `Host` header. We should wait until we get @whatyouhide's thoughts on https://github.com/elixir-mint/mint/pull/283#issuecomment-717376133.

myobie:
> I don't think we should add a `:hostname` transport option that only the SSL transport uses and does not affect the `Host` header. We should wait until we get @whatyouhide's thoughts on [#283 (comment)](https://github.com/elixir-mint/mint/pull/283#issuecomment-717376133).

OK, I was thinking the same thing. I wanted to first get a test to pass to show that `:ssl` does support the unix socket. I think having a `:hostname` option that could then be merged into `:transport_opts` so it could be used for both the `Host` header and for the ssl is very doable and I can get that working tomorrow or next week.

whatyouhide:
@ericmj I start to see the value in renaming all docs to have `address` in place of `hostname`, and adding a `hostname` option that sets the value of the `Host` header (with sensible defaults so that for 99% of use cases users don't have to set it).

ericmj:
Great, thanks for the feedback. So I think these are the steps we should do:

1. Rename `hostname` parameter to `address` and accept all address types the underlying transports accept.
2. Add a `:hostname` connection option. This option will be used for the `Host` header, and passed down as an option to the transports to use for hostname verification, SNI, etc.
3. The `:hostname` option is required if `address` is not a string, otherwise it defaults to the same value as `address`.

This should hopefully address (no pun intended) all concerns in this PR and in #282. WDYT?

myobie:
I've updated the top comment / description to include the steps you mentioned @ericmj. Thanks for being open to the change @whatyouhide.

myobie:
@ericmj I've pushed my first attempt at accepting all types of addresses along with a new `:hostname` option. I currently place a `:hostname` option into `transport_opts` every time, so it will either be the address or the `:hostname` option provided during connect. The `Host` header is being set like it was before, since I didn't modify the conn structs – we don't need the `address` anymore after `connect` makes a `socket`.

I also noticed the tests for OTP 18.3 + elixir 1.53 are failing because `erlex` now requires `>= 1.6`. Should we remove that version combo from the matrix or mark that test as optional?

ericmj:
> I also noticed the tests for OTP 18.3 + elixir 1.53 are failing because erlex now requires >= 1.6. Should we remove that version combo from the matrix or mark that test as optional?

I don’t see any tests failing because of erlex, I only see the new tests failing. Can you link to the failed test?

test/mint/unix_socket_test.exs
```diff
@@ -0,0 +1,51 @@
+defmodule Mint.UnitSocketTest do
+  use ExUnit.Case, async: true
+
+  alias Mint.{HTTP, TestSocketServer}
+
+  require HTTP
+
+  test "starting an http connection to a unix domain socket works" do
+    unless is_unix?() && otp_19?() do
```
ericmj, line 9: Let’s change this to a `@tag :skip` before the test.

test/support/mint/http1/test_server.ex
```diff
@@ -1,5 +1,5 @@
 defmodule Mint.HTTP1.TestServer do
-  def start() do
+  def start do
```
ericmj, line 2: Let’s revert this change.

test/support/mint/http1/test_server.ex
```diff
@@ -11,7 +11,7 @@ defmodule Mint.HTTP1.TestServer do
     end
   end
 
-  defp loop(listen_socket, parent, server_ref) do
+  def loop(listen_socket, parent, server_ref) do
```
ericmj, line 14: Let’s revert this change.

test/mint/http1/conn_test.exs
```diff
@@ -1,7 +1,8 @@
 defmodule Mint.HTTP1Test do
   use ExUnit.Case, async: true
 
-  alias Mint.{HTTPError, HTTP1, HTTP1.TestServer}
+  alias Mint.{HTTPError, HTTP1}
+  alias HTTP1.TestServer
```
ericmj, line 5: Let’s revert this change.

ericmj (commented):
Looks great, I only have a few comments.

myobie:
> > I also noticed the tests for OTP 18.3 + elixir 1.53 are failing because erlex now requires >= 1.6. Should we remove that version combo from the matrix or mark that test as optional?
> 
> I don’t see any tests failing because of erlex, I only see the new tests failing. Can you link to the failed test?

Gah, yes, you are right. I misunderstood this error https://github.com/elixir-mint/mint/pull/283/checks?check_run_id=1351815687#step:6:12

lib/mint/core/transport/ssl.ex
```diff
@@ -311,25 +311,35 @@ defmodule Mint.Core.Transport.SSL do
   #       crl_cache: {:ssl_crl_cache, {:internal, [http: 30_000]}}
 
   @impl true
-  def connect(hostname, port, opts) do
-    hostname = String.to_charlist(hostname)
+  def connect(address, port, opts) do
+    hostname = opts |> Mint.Core.Util.hostname(address)
+    opts = Keyword.delete(opts, :hostname)
```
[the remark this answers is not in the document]
myobie: It does complain and we see a lot of test failures with `%Mint.TransportError{reason: {:options, {:socket_options, [hostname: "localhost", ...]}}}` in the output

test/mint/unix_socket_test.exs
myobie: @ericmj OK, I've made a helper to conditionally tag it as `:skip` and I've added `:skip` to the list to be excluded by default.

test/test_helper.exs
```diff
@@ -1,3 +1,3 @@
-ExUnit.start(exclude: :proxy)
+ExUnit.start(exclude: [:proxy, :skip])
```
ericmj, line 1: This change is not necessary.

ericmj (approved):
This looks really good. Just one final comment.

whatyouhide (changes requested):

lib/mint/types.ex
```diff
@@ -3,6 +3,12 @@ defmodule Mint.Types do
   HTTP-related types.
   """
 
+  @typedoc """
+  A hostname, ip address, unix domain socket path, :loopback, or any
+  other term representing an internet address.
+  """
+  @type address :: :inet.socket_address() | :inet.hostname()
```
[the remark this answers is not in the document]
ericmj: Good catch. Maybe we should replace `:inet.hostname()` with `String.t()`?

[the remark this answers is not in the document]
myobie: [`:inet.socket_address()`](http://erlang.org/doc/man/inet.html#type-socket_address) allows `any` so it does allow `binary` technically, but I think it could be a good idea to replace these with a more explicit list of "desired types" instead of the list of "technically allowed types" by the underlying `connect` functions.

lib/mint/types.ex
```diff
@@ -3,6 +3,12 @@ defmodule Mint.Types do
   HTTP-related types.
   """
 
+  @typedoc """
+  A hostname, ip address, unix domain socket path, :loopback, or any
+  other term representing an internet address.
+  """
+  @type address() :: :inet.socket_address() | :inet.hostname()
```
myobie, line 10:
```suggestion
  @type address() :: :inet.socket_address() | String.t()
```

Related to https://github.com/elixir-mint/mint/pull/283#discussion_r518203488 

Also I misunderstood the erlang `any` to be the same as the elixir typespec `any` and I just realized it would actually be the atom `:any` so I was wrong with my most recent comment in the previous discussion 😓 sorry about that. I will say going back and forth from erlang to elixir every day the past couple weeks has been tough 😅 

/cc @whatyouhide and @ericmj

whatyouhide (approved):

PR #307, opened 2021-04-03
Remove `@doc since:` for pre-1.0.0 functions

wojtekmach:
Given Mint reached the 1.0 milestone, I don't think `@doc since` on pre-1.0 additions is useful but of course it doesn't hurt either.


ericmj (approved):

whatyouhide:
Thanks @wojtekmach, good catch and yep good point :)

PR #321, opened 2021-05-11
support SETTINGS_ENABLE_CONNECT_PROTOCOL server-side setting

the-mikedavis:
Hello! :wave:

I'm a huge fan of Mint, thank y'all for maintaining it :slightly_smiling_face:

I've been eyeing [#106](https://github.com/elixir-mint/mint/issues/106) lately and I've been able to get HTTP/1 websockets going with just some minor hacks in [`NFIBrokerage/mint_web_socket`](https://github.com/NFIBrokerage/mint_web_socket). The HTTP/2 bootstrap extends rfc7540's SETTINGS parameters, though, so it looks like we'll need to some changes here to support HTTP/2.

Specifically [rfc8441](https://www.rfc-editor.org/rfc/rfc8441.html) adds a settings param SETTINGS_ENABLE_CONNECT_PROTOCOL with [a new code of `0x8`](https://www.rfc-editor.org/rfc/rfc8441.html#section-9.1) defaulting to `0`.

This PR adds a new `t:Mint.HTTP2.setting/0` `:enable_connect_protocol` defaulting to `false` and the encoding and decoding clauses in `Mint.HTTP2.Frame`.

### DIFF
--- a/lib/mint/http2.ex
+++ b/lib/mint/http2.ex
@@ -197,7 +197,8 @@ defmodule Mint.HTTP2 do
       max_concurrent_streams: 100,
       initial_window_size: @default_window_size,
       max_frame_size: @default_max_frame_size,
-      max_header_list_size: :infinity
+      max_header_list_size: :infinity,
+      enable_connect_protocol: false
     },
 
     # Settings that the client communicates to the server.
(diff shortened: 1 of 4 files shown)


lib/mint/http2.ex
```diff
@@ -197,7 +197,8 @@ defmodule Mint.HTTP2 do
       max_concurrent_streams: 100,
       initial_window_size: @default_window_size,
       max_frame_size: @default_max_frame_size,
-      max_header_list_size: :infinity
+      max_header_list_size: :infinity,
+      enable_connect_protocol: false
```
the-mikedavis, line 201: I believe this is a server-only setting. The rfc has a paragraph [in section 3](https://www.rfc-editor.org/rfc/rfc8441.html#section-3):

> Upon receipt of SETTINGS_ENABLE_CONNECT_PROTOCOL with a value of 1, a client MAY use the Extended CONNECT as defined in this document when creating new streams.  Receipt of this parameter by a server does not have any impact.

ericmj (approved):

whatyouhide (approved):
Nice work! 💯

PR #344, opened 2021-12-22
Doc update dead links to RFC 7540

slashmili:
The links to http2-spec are giving 404. 
I used the link to https://httpwg.org/specs/rfc7540.html as oppose to https://datatracker.ietf.org/doc/html/rfc7540 because it's used in https://http2.github.io main page.


whatyouhide (approved):
Good catch, thanks @slashmili! 💟

PR #350, opened 2022-04-27
Fix compilation warnings around CAStore

whatyouhide:
* Mix was complaining when compiling Mint as a dependency if `:castore` was not in the dependencies too. Solved this by not using `CAstore` as a compile-time module reference, but making it dynamic instead.
* **Unrelated refactor**: I just used `Keyword.put_new/3` and `Keyword.put_new_lazy/3` in favor of some `if` checks in `Mint.Core.Transport.SSL`.

### DIFF
--- a/lib/mint/core/transport/ssl.ex
+++ b/lib/mint/core/transport/ssl.ex
@@ -475,24 +475,14 @@ defmodule Mint.Core.Transport.SSL do
   end
 
   defp add_customize_hostname_check(opts) do
-    customize_hostname_check_present? = Keyword.has_key?(opts, :customize_hostname_check)
-
-    if not customize_hostname_check_present? do
-      Keyword.put(opts, :customize_hostname_check, match_fun: &match_fun/2)
-    else
-      opts
-    end
+    Keyword.put_new(opts, :customize_hostname_check, match_fun: &match_fun/2)
   end
 
   defp add_verify_fun(opts, host_or_ip) do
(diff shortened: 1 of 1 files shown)


lib/mint/core/transport/ssl.ex
```diff
@@ -565,11 +555,14 @@ defmodule Mint.Core.Transport.SSL do
   end
 
   defp add_cacerts(opts) do
-    if Keyword.has_key?(opts, :cacertfile) || Keyword.has_key?(opts, :cacerts) do
+    if Keyword.has_key?(opts, :cacertfile) or Keyword.has_key?(opts, :cacerts) do
```
whatyouhide, line 558: Unrelated refactor, strict `or` since both sides are boolean.

lib/mint/core/transport/ssl.ex
```diff
@@ -565,11 +555,14 @@ defmodule Mint.Core.Transport.SSL do
...
       opts
     else
       raise_on_missing_castore!()
-      Keyword.put(opts, :cacertfile, CAStore.file_path())
+      # Use a dynamic module name to avoid compilation warnings when Mint is used as
+      # a dependency without :castore.
+      castore_mod = CAStore
+      Keyword.put(opts, :cacertfile, castore_mod.file_path())
```
whatyouhide, line 565: Main fix.

ericmj (approved):

PR #354, opened 2022-05-20
Mention OTP25+ cert management in errors and docs

whatyouhide:
cc @ericmj @josevalim

### DIFF
--- a/lib/mint/core/transport/ssl.ex
+++ b/lib/mint/core/transport/ssl.ex
@@ -692,10 +692,13 @@ defmodule Mint.Core.Transport.SSL do
     Code.ensure_loaded?(CAStore) ||
       raise """
       default CA trust store not available; please add `:castore` to your project's \
-      dependencies or specify the trust store using the :cacertfile/:cacerts option
-      within :transport_options. From OTP 25, you can also use :public_key.cacerts_get/0
-      to get certificates that you loaded from files or from the OS with \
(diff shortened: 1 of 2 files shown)


lib/mint/core/transport/ssl.ex
```diff
@@ -690,9 +690,13 @@ defmodule Mint.Core.Transport.SSL do
...
-      raise "default CA trust store not available; " <>
-              "please add `:castore` to your project's dependencies or " <>
-              "specify the trust store using the `:cacertfile`/`:cacerts` option"
+      raise """
+      default CA trust store not available; please add `:castore` to your project's \
+      dependencies or specify the trust store using the :cacertfile/:cacerts option
+      within :transport_options. From OTP 25, you can also use :public_key.cacerts_get/0
+      to get certificates that you loaded from files or from the OS with \
+      :public_key.cacerts_load/0,1. See: https://www.erlang.org/blog/my-otp-25-highlights/#ca-certificates-can-be-fetched-from-the-os-standard-place
+      """
```
josevalim, line 699: Shall we use a bulleted list with two items? :)

josevalim (approved):

lib/mint/core/transport/ssl.ex
whatyouhide: Yep, done!

PR #355, opened 2022-05-20
Correct `get_window_size/2` docs

daskycodes:
This PR corrects the documentation for the `HTTP2.get_window_size/2` function.

This function only exists in the `HTTP2` module, and pattern matches on the `:request` atom instead of `:request_ref`

https://github.com/elixir-mint/mint/blob/e7199d1163e5a8a1b0001cacc93063e866dbaa35/lib/mint/http2.ex#L759-L768

### DIFF
--- a/lib/mint/http2.ex
+++ b/lib/mint/http2.ex
@@ -735,17 +735,17 @@ defmodule Mint.HTTP2 do
 
   On the connection:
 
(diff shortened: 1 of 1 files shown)


whatyouhide:
Ah, great catches! Thanks @daskycodes ❤️

PR #358, opened 2022-05-30
Add CAStore to xref excludes

wojtekmach:


mix.exs
```diff
@@ -17,7 +17,8 @@ defmodule Mint.MixProject do
       xref: [
         exclude: [
           :persistent_term,
-          {:ssl, :cipher_suites, 1}
+          {:ssl, :cipher_suites, 1},
+          CAStore
```
whatyouhide, line 21: Noob question: won't this still warn if someone uses Mint as dependency?

wojtekmach: nope!

You can see it in action here:

```
$ elixir -e 'Mix.install([{:req, github: "wojtekmach/req", ref: "f0a59ef"}])'
$ elixir -e 'Mix.install([{:req, github: "wojtekmach/req", ref: "caeefae"}])'
```

([req#caeefae](https://github.com/wojtekmach/req/commit/caeefaea471c8e864287b4cff4c259c80ae35e68))

whatyouhide:
Thanks @wojtekmach! 💟

PR #363, opened 2022-06-13
Add support for interim responses in HTTP/2

### LINKED ISSUE #349: Setting user-agent header leads to protocol error
Hello,

It seems that when I set `user-agent` header to some specific values, requests return an error message about `:protocol_error`

I'm using mint `1.4.1` and a genserver module taken from your [docs](https://hexdocs.pm/mint/architecture.html#wrapping-a-mint-connection-in-a-genserver).

This works fine:
```
{:ok, pid} = ConnectionProcess.start_link({:https, "www.shopify.com", 443})
response = ConnectionProcess.request(pid, "GET", "/", [{"user-agent", "whatever"}], :nil)
```

But this throws an error:
```
response = ConnectionProcess.request(pid, "GET", "/", [{"user-agent", "curl/7.68.0"}], :nil)

%Mint.HTTPError{module: Mint.HTTP2, reason: {:protocol_error, "trailing headers didn't set the END_STREAM flag"}}}
```

Getting this url with `curl` works OK.

whatyouhide:
Closes #349.

### DIFF
--- a/lib/mint/http.ex
+++ b/lib/mint/http.ex
@@ -675,6 +675,8 @@ defmodule Mint.HTTP do
 
     * `{:status, request_ref, status_code}` - returned when the server replied
       with a response status code. The status code is a non-negative integer.
+      You can have zero or more `1xx` `:status` responses for a single request,
+      but they all precede a single non-`1xx` `:status` response.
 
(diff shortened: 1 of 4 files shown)


ericmj (approved):
💜

PR #367, opened 2022-06-30
Properly close socket on erroneous connection

### LINKED ISSUE #365: close/1 does not close socket properly on externally closed connection
Hello, guys!

I was testing my application to work in extremely edge cases and noticed that mint does not close sockets correctly under certain circumstances, this leads to a leak.

Mint does not close a socket when two circumstances coincide:
1. Connection was closed externally, for example, by timeout
2. Your process for some reason does not process ssl_closed message

I wrote simple test to demonstrate this
```elixir
 @tag timeout: :infinity
  test "close/1 does not close socket properly on externally closed connection" do
    # Without timeout
    assert {:ok, %HTTP2{} = conn1} = HTTP2.connect(:https, "nghttp2.org", 443)
    IO.puts "Opened first connection ✓"

    assert {:ok, conn1} = HTTP2.close(conn1)
    IO.puts "Closed first connection ✓"

    refute HTTP2.open?(conn1)
    IO.puts "Connection closed successfully ✓"

    refute (conn1.socket |> elem(2) |> Enum.all?(fn pid -> Process.alive?(pid) end))
    IO.puts "Connection socket's processes died sucessfully ✓"


    # With timeout
    assert {:ok, %HTTP2{} = conn2} = HTTP2.connect(:https, "nghttp2.org", 443)
    IO.puts "Opened second connection ✓"

    IO.puts "Sleep 5 minutes zzzZZ"
    :timer.sleep(300_000) # yes, I know that we should listen messages and update connection, but wi ar bad

    assert {:error, conn2, %Mint.TransportError{reason: :closed}} = Mint.HTTP.request(conn2, "GET", "/", [], "")
    IO.puts "Tried to request with this connection and it ends with Mint.TransportError, connection dead for sure"

    assert {:ok, conn2} = HTTP2.close(conn2)
    IO.puts "Closed second connection ✓"

    refute HTTP2.open?(conn2)
    IO.puts "Connection closed successfully ✓"

    refute (conn2.socket |> elem(2) |> Enum.all?(fn pid -> Process.alive?(pid) end)), "Connection socket's processes not died :("
  end
```

Test results:
mint_bug

I quickly found the error by looking at the source code:
1. When `close/1` called we are here https://github.com/elixir-mint/mint/blob/7aaf0377343f63dca87d8007fd404253a44598b5/lib/mint/http2.ex#L387
2. Then move on to `send_connection_error!/3` https://github.com/elixir-mint/mint/blob/7aaf0377343f63dca87d8007fd404253a44598b5/lib/mint/http2.ex#L2027
3. And then move on to `send!/2` https://github.com/elixir-mint/mint/blob/7aaf0377343f63dca87d8007fd404253a44598b5/lib/mint/http2.ex#L2094

```elixir
  defp send!(%Mint.HTTP2{transport: transport, socket: socket} = conn, bytes) do
    case transport.send(socket, bytes) do
      :ok ->
        conn

      {:error, %TransportError{reason: :closed} = error} ->
        throw({:mint, %{conn | state: :closed}, error})

      {:error, reason} ->
        throw({:mint, conn, reason})
    end
  end
```

And there are problem, `transport.send(socket, bytes)` returns `TransportError` then you just throw this error and because of it we ignore this command `_ = conn.transport.close(conn.socket)` at `send_connection_error!/3` and the socket is not closed correctly

```elixir
  defp send_connection_error!(conn, error_code, debug_data) do
    frame =
      goaway(stream_id: 0, last_stream_id: 2, error_code: error_code, debug_data: debug_data)

    _ = conn.transport.close(conn.socket) # <<<--- this line ignored because send! explodes
    conn = put_in(conn.state, :closed)
    throw({:mint, conn, wrap_error({error_code, debug_data})})
  end
```

I hope I made my point clear. Looking at the code, I see that you wanted to handle such extreme edge cases, but something went wrong 😂

bitberry-dev:
Fix #365

### DIFF
--- a/lib/mint/http2.ex
+++ b/lib/mint/http2.ex
@@ -2028,10 +2028,11 @@ defmodule Mint.HTTP2 do
     frame =
       goaway(stream_id: 0, last_stream_id: 2, error_code: error_code, debug_data: debug_data)
 
-    conn = send!(conn, Frame.encode(frame))
+    # try send goaway frame and close connection
+    _ = conn.transport.send(conn.socket, Frame.encode(frame))
     _ = conn.transport.close(conn.socket)
-    conn = put_in(conn.state, :closed)
-    throw({:mint, conn, wrap_error({error_code, debug_data})})
+
+    throw({:mint, %{conn | state: :closed}, wrap_error({error_code, debug_data})})
   end
 
   defp close_stream!(conn, stream_id, error_code) do
--- a/lib/mint/http2.ex
+++ b/lib/mint/http2.ex
@@ -2028,7 +2028,9 @@ defmodule Mint.HTTP2 do
     frame =
       goaway(stream_id: 0, last_stream_id: 2, error_code: error_code, debug_data: debug_data)
 
-    # try send goaway frame and close connection
+    # Try to send the GOAWAY frame and close connection.
+    # If the frame fails to send, we still want to set the close
+    # the socket, set the connection state to :closed, and return an error.
     _ = conn.transport.send(conn.socket, Frame.encode(frame))
     _ = conn.transport.close(conn.socket)
 
--- a/test/mint/http2/conn_test.exs
+++ b/test/mint/http2/conn_test.exs
@@ -2,6 +2,7 @@ defmodule Mint.HTTP2Test do
   use ExUnit.Case, async: true
 
   import Mint.HTTP2.Frame
+  import Mint.HTTP2.TestHelpers, only: [extract_port: 1]
   import ExUnit.CaptureLog
 
   alias Mint.{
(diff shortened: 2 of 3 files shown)


test/mint/http2/conn_test.exs
```diff
@@ -391,6 +391,35 @@ defmodule Mint.HTTP2Test do
...
+      assert {:error, _} = :ssl.getstat(socket)
+    end
+
+    test "close/1 properly closes socket on errornous connection", %{conn: conn} do
+      # force the transport to one that always times out on send
+      conn = %{conn | transport: Mint.HTTP2.TestTransportSendTimeout}
+
+      # Check socket status, before close it should be :ok
+      socket = HTTP2.get_socket(conn)
+      assert {:ok, _} = :ssl.getstat(socket)
```
whatyouhide, line 414: `ssl:getstat/1` was introduced in OTP 19, but we're still supporting OTP 18 in Mint. Could you try using `inet:getstat/1` here? I believe that one was there earlier.

lib/mint/http2.ex
```diff
@@ -2028,10 +2028,11 @@ defmodule Mint.HTTP2 do
...
       goaway(stream_id: 0, last_stream_id: 2, error_code: error_code, debug_data: debug_data)
 
-    conn = send!(conn, Frame.encode(frame))
+    # try send goaway frame and close connection
+    _ = conn.transport.send(conn.socket, Frame.encode(frame))
     _ = conn.transport.close(conn.socket)
-    conn = put_in(conn.state, :closed)
-    throw({:mint, conn, wrap_error({error_code, debug_data})})
+
+    throw({:mint, %{conn | state: :closed}, wrap_error({error_code, debug_data})})
```
whatyouhide, line 2035: Why the change here from `put_in/2` to `%{conn | ...}`?

lib/mint/http2.ex
```diff
@@ -2028,10 +2028,11 @@ defmodule Mint.HTTP2 do
     frame =
       goaway(stream_id: 0, last_stream_id: 2, error_code: error_code, debug_data: debug_data)
 
-    conn = send!(conn, Frame.encode(frame))
+    # try send goaway frame and close connection
```
whatyouhide, line 2031:
```suggestion
    # Try to send the GOAWAY frame and close connection.
    # If the frame fails to send, we still want to set the close
    # the socket, set the connection state to :closed, and return an error.
```

lib/mint/http2.ex
bitberry-dev: Differences occur only when state key is missing in conn. I decided to make same behaviour as was in failure scenario in `send!/2` https://github.com/elixir-mint/mint/blob/90099586704e7c6e2e190c5b127b7db9a36cffe8/lib/mint/http2.ex#L2112

Was there any reason to use `put_in/2`? Did this method require conn to have a state?

test/mint/http2/conn_test.exs
bitberry-dev: Oh, I didn't know, thanks, I'll check what options are available for ssl socket in OTP 18.

whatyouhide (approved):

epinault:
Hmm. I am seeing issues with Finch/Mint around socket closed that I don't see when using other things like hackney with Tesla . 

I was thinking a service we run had issues but even so, hackney is ok with it but with Finch/Mint I see some socket closed happening .. I am wondering if there is a path still where that can happen

PR #375, opened 2022-08-17
Check argument types on `HTTP1.request/5`

britto:
The same way we do in `HTTP2.request/5`. Noticed it through dialyzer while passing a map of `headers` instead of a list.


ericmj (approved):

ericmj (approved):

whatyouhide:
Thanks @britto 💟

Issue #384, opened 2023-02-22
HTTP/2 server settings are handled as client settings


hissssst:
## Overview

Setting server settings with `put_settings` fails, while the HTTP2 spec describes that all settings are negotiable.

## Environment

```sh
$ uname -srm
Linux 5.15.90 x86_64

$ elixir --version
Erlang/OTP 25 [erts-13.1.4] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit:ns]

Elixir 1.14.3 (compiled with Erlang/OTP 25)

$ git log | head -n5
commit 1e70ccffbcce6f4d601661f22a700e001ce8e475
Author: Andrea leopardi <an.leopardi@gmail.com>
Date:   Mon Jun 13 14:29:25 2022 +0200

    Release v1.4.2
```

## Reproduction

```elixir
alias Mint.HTTP2

{:ok, conn} = HTTP2.connect(:https, "example.com", 443)
{:ok, conn} = HTTP2.put_settings(conn, initial_window_size: 2097152)
{:ok, conn, request_ref} = HTTP2.request(conn, "GET", "/", [], "")

msg = receive do: (msg -> msg)
{:ok, conn, _} = HTTP2.stream(conn, msg)

msg = receive do: (msg -> msg)
{:ok, conn, _} = HTTP2.stream(conn, msg)
```

## Expected beavior

Option is set, no errors occurred

## Actual behavior

```elixir
** (FunctionClauseError) no function clause matching in anonymous fn/2 in Mint.HTTP2.apply_client_settings/2

    The following arguments were given to anonymous fn/2 in Mint.HTTP2.apply_client_settings/2:

        # 1
        {:initial_window_size, 2097152}

        # 2
        %Mint.HTTP2{
          transport: Mint.Core.Transport.SSL,
          socket: {:sslsocket, {:gen_tcp, #Port<0.7>, :tls_connection, :undefined},
           [#PID<0.276.0>, #PID<0.275.0>]},
          mode: :active,
          hostname: "example.com",
          port: 443,
          scheme: "https",
          state: :open,
          buffer: "",
          window_size: 1048576,
          encode_table: %HPAX.Table{
            max_table_size: 4096,
            entries: [],
            size: 0,
            length: 0
          },
          decode_table: %HPAX.Table{
            max_table_size: 4096,
            entries: [],
            size: 0,
            length: 0
          },
          ping_queue: {[], []},
          client_settings_queue: {[], []},
[...]

PR #386, opened 2023-02-26
Support all possible HTTP/2 client settings

whatyouhide:
Closes https://github.com/elixir-mint/mint/issues/384.

/cc @hissssst


lib/mint/http2.ex
```diff
@@ -199,6 +207,8 @@ defmodule Mint.HTTP2 do
     # Settings that the client communicates to the server.
     client_settings: %{
```
whatyouhide, line 208: I want to make this a struct maybe, so that we have more precision in the fields. It's just a tiny `__struct__` field more, so I don't think it matters much. It's an internal detail also, since these are not exposed.

lib/mint/http2.ex
```diff
@@ -1274,14 +1284,14 @@ defmodule Mint.HTTP2 do
...
-    validate_settings!(settings)
+    validate_client_settings!(settings)
     frame = settings(stream_id: 0, params: settings)
     conn = send!(conn, Frame.encode(frame))
     conn = update_in(conn.client_settings_queue, &:queue.in(settings, &1))
     conn
   end
 
-  defp validate_settings!(settings) do
+  defp validate_client_settings!(settings) do
```
whatyouhide, line 1294: Renamed because this is only for client settings.

PR #404, opened 2023-06-23
Correct architecture link in Decompression.md

zorbash:
The correct URL seems to be https://github.com/elixir-mint/mint/blob/main/pages/Architecture.md#content


whatyouhide:
Thanks @zorbash! ❤️

PR #434, opened 2024-04-30
Don't send RST_STREAM when not needed

### LINKED ISSUE #433: Avoid unnecessary `RST_STREAM` frames in HTTP/2 implementation
When a stream is locally closed and a EndStream flag is received, Mint is sending RST_STREAM with NO_ERROR.
This is unnecessary as a stream can be deemed closed by both sides at this point, and this frame is just ignored.

v0idpwn:
When the stream state is local_half_closed and an end_stream flag is received, the stream can be terminated normally without RST_STREAM.

Closes #433.

I still didn't work on integration tests, and while current tests kind of cover this feature, I'll add a specific test for it.

### DIFF
--- a/lib/mint/http2.ex
+++ b/lib/mint/http2.ex
@@ -1566,7 +1566,7 @@ defmodule Mint.HTTP2 do
         responses = [{:data, stream.ref, data} | responses]
 
         if flag_set?(flags, :data, :end_stream) do
-          conn = close_stream!(conn, stream.id, :no_error)
+          conn = close_stream!(conn, stream.id, :remote_end_stream)
           {conn, [{:done, stream.ref} | responses]}
         else
           {conn, responses}
@@ -1675,7 +1675,7 @@ defmodule Mint.HTTP2 do
                 {conn, responses}
 
               end_stream? ->
-                conn = close_stream!(conn, stream.id, :no_error)
+                conn = close_stream!(conn, stream.id, :remote_end_stream)
                 {conn, [{:done, ref} | new_responses]}
 
               true ->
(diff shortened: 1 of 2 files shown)


v0idpwn:
This is a WIP, I'm opening for reference. I need to double check some corner cases and maybe my implementation can be simplified.

lib/mint/http2.ex
```diff
@@ -2094,18 +2094,27 @@ defmodule Mint.HTTP2 do
...
+      cond do
+        # If the stream is ended on both sides, it is already deemed closed and
+        # there's no need to send a RST_STREAM frame
+        reason == :remote_end_stream and stream.state == :half_closed_local ->
+          conn
+
+        # We send a RST_STREAM with the given error code so that we move the
+        # stream to the :closed state (that is, we remove it).
+        open?(conn) ->
+          error_code = if reason == :remote_end_stream, do: :no_error, else: reason
```
v0idpwn, line 2111: I don't think this line is ever used, but it doesn't feel right to remove it (?)

v0idpwn (commented):
I'm not 100% happy with the implementation because I think it's a bit dirty, but it solves the problem and is pretty small, so I'm opening the PR.

v0idpwn:
Also it looks like CI is failing for totally unrelated reasons™️

whatyouhide:
> Also it looks like CI is failing for totally unrelated reasons™️

Yes Totally Unrelated™, no worries

> I'm not 100% happy with the implementation because I think it's a bit dirty, but it solves the problem and is pretty small, so I'm opening the PR.

@v0idpwn is this ready to review?

v0idpwn:
Yes!

whatyouhide:
Woot woot!

PR #435, opened 2024-06-08
Default to using Erlang certificates store

josevalim:
The OTP team no longer supports Erlang versions earlier than 25+, so we can assuming that `:public_key.cacerts_get/0` is available and only fallback to `CAStore` if not.

This also solves a bug in that Req/Finch/Mint do not work inside escripts by default (because inside an escript you cannot access the priv dir of an application).

### DIFF
--- a/lib/mint/core/transport/ssl.ex
+++ b/lib/mint/core/transport/ssl.ex
@@ -572,8 +572,13 @@ defmodule Mint.Core.Transport.SSL do
     if Keyword.has_key?(opts, :cacertfile) or Keyword.has_key?(opts, :cacerts) do
       opts
     else
-      raise_on_missing_castore!()
-      Keyword.put(opts, :cacertfile, CAStore.file_path())
+      try do
+        Keyword.put(opts, :cacerts, :public_key.cacerts_get())
+      rescue
+        _ ->
+          raise_on_missing_castore!()
+          Keyword.put(opts, :cacertfile, CAStore.file_path())
+      end
     end
   end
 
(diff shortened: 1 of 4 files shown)


josevalim:
It would be fantastic to get a new minor version with this change to address this: https://github.com/livebook-dev/livebook/issues/2642

But we can also depend on main if merged. :)

lib/mint/core/transport/ssl.ex
```diff
@@ -572,8 +572,13 @@ defmodule Mint.Core.Transport.SSL do
     if Keyword.has_key?(opts, :cacertfile) or Keyword.has_key?(opts, :cacerts) do
       opts
     else
-      raise_on_missing_castore!()
-      Keyword.put(opts, :cacertfile, CAStore.file_path())
+      try do
+        Keyword.put(opts, :cacerts, :public_key.cacerts_get())
+      rescue
```
whatyouhide, line 577: Should we catch `UndefinedFunctionError` specifically?

wojtekmach: I believe it is for when running OTP 25 but OS happens not to have certa, cacerts_get() raises then.

whatyouhide:
@josevalim 1.6.1 is out 🙃

wojtekmach:
@whatyouhide v1.6.1 is out on GitHub but looks like not yet on Hex!

wojtekmach:
Thanks! I'm updating Finch accordingly: https://github.com/sneako/finch/pull/274.

PR #441, opened 2024-07-01
Fix incorrect CI badge which still point to Travis CI

kianmeng:
This should switch the CI badge GitHub badge.


kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #442, opened 2024-07-01
Fix cannot build Docker images using Podman Compose

kianmeng:
This PR resolves the following error when using `podman-compose`:

    Error: short-name "caddy:2.6.4-alpine" did not resolve to an alias and
    no unqualified-search registries are defined in
    "/etc/containers/registries.conf"

We also bump the Docker image for Caddy and Alpine Linux.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,12 +12,12 @@ services:
       - "8889:8888"
 
   httpbin:
-    image: kennethreitz/httpbin:latest
+    image: docker.io/kennethreitz/httpbin:latest
     ports:
       - "8080:80"
 
   caddyhttpbin:
-    image: caddy:2.6.4-alpine
+    image: docker.io/caddy:2.8.4-alpine
     # In GitHub Actions we want to access the files created
     # by Caddy. However, in Linux these end up being owned by root
     # because the container by default runs using the root user
--- a/test/docker/tinyproxy-auth/Dockerfile
+++ b/test/docker/tinyproxy-auth/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.17.2
+FROM docker.io/alpine:3.20.1
 
 RUN apk add --no-cache \
 	tinyproxy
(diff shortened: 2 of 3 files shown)


kianmeng:
@whatyouhide Not sure why, I kept getting this two test cases that failed on me, but these tests is pointing to `localhost`?

```
podman-compose up --force-recreate -d
mix test
```

```
  1) test connect/3 does not fall back to IPv4 if IPv4 is disabled (Mint.Core.Transport.TCPTest)
     test/mint/core/transport/tcp_test.exs:63
     match (=) failed
     code:  assert {:error, %Mint.TransportError{reason: :econnrefused}} =
              TCP.connect("localhost", port, active: false, inet6: true, inet4: false, timeout: 1
000)            
     left:  {:error, %Mint.TransportError{reason: :econnrefused}}
     right: {:error, %Mint.TransportError{reason: :nxdomain, __exception__: true}}
     stacktrace:                                                                                 
       test/mint/core/transport/tcp_test.exs:80: (test)                                          
                                                                                                 
..
  2) test connect/3 does not fall back to IPv4 if IPv4 is disabled (Mint.Core.Transport.SSLTest)      test/mint/core/transport/ssl_test.exs:250
     match (=) failed
     code:  assert {:error, %Mint.TransportError{reason: :econnrefused}} =
              SSL.connect("localhost", port,
                active: false,
                inet6: true,
                inet4: false,
                timeout: 1000,
                verify: :verify_none
              )
     left:  {:error, %Mint.TransportError{reason: :econnrefused}}
     right: {:error, %Mint.TransportError{reason: :nxdomain, __exception__: true}}
     stacktrace:
       test/mint/core/transport/ssl_test.exs:269: (test)
```

kianmeng:
> @kianmeng do you get them with Docker too? Because they are not failing in CI, it looks like.

I disabled the DNS adblocker, which seemed to work. Not sure why the adblocker was aggressively blocking localhost as well.

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

Issue #437, opened 2024-06-13
Add `humanize_error_code` generic case


kaiserdan:
The humanize_error_code function should have a generic case for error codes that aren't explicitly defined.  I sometimes get a 70007 error which then causes an error in the module. Happy to make a PR for this if you like. 

```
(mint 1.6.0) lib/mint/http2.ex:823: Mint.HTTP2.stream/2
(mint 1.6.0) lib/mint/http2.ex:1397: Mint.HTTP2.maybe_concat_and_handle_new_data/2
(mint 1.6.0) Lib/mint/http2. ex:1402: Mint.HTTP2.handle_new_data/3
(mint 1.6.0) lib/mint/http2/frame.ex:99: Mint.HTTP2.Frame.decode_next/2
(mint 1.6.0) lib/mint/http2/frame.ex:237: Mint.HTTP2.Frame.decode_goaway/3
(mint 1.6.0) lib/mint/http2/frame.ex:472: Mint.HTTP2.Frame.humanize_error_code (70007)
** (FunctionClauseError) no function clause matching in Mint.HTTP2.Frame.humanize_error_code/1
```

whatyouhide:
@kaiserdan yes would love the help. Does the HTTP/2 spec say what to do in case of unknown errors?

kaiserdan:
@whatyouhide I don't think it's part of the HTTP spec but rather a custom error code thats generated by some webservers or proxies

I solved it locally by adding a generic case for it it as `defp humanize_error_code(_), do: :internal_error `

Maybe a dedicated error code makes sense like : undefined_error, :unknown_error, :other_error, or :custom_error ?

Let me know what you prefer!

whatyouhide:
Yeah I would like something like `{:custom_error, code}`. Thoughts? You're welcome to start on a PR and we can continue the discussion there 🙃

PR #445, opened 2024-07-02
Add generic :custom_error case to humanize_error_code.

kaiserdan:
Sometimes webservers respond with custom error codes which causes a FunctionClauseError in the humanize_error_code function. This change adds a generic case of the function that handles those cases.

Closes https://github.com/elixir-mint/mint/issues/437.


whatyouhide:
Nice! We will need a test for this, can you add one?

kaiserdan:
Does that look good, or is there a better way to test it?

PR #447, opened 2024-07-26
Add Mint.HTTP.recv_response/3

wojtekmach:
I'd like to propose to add a `Mint.HTTP1.recv_response/3` that improve ergonomics of making one off requests:

```elixir
iex> {:ok, conn} = Mint.HTTP.connect(:https, "httpbin.org", 443, mode: :passive)
iex> {:ok, conn, ref} = Mint.HTTP.request(conn, "GET", "/user-agent", [], nil)
iex> {:ok, conn, response} = Mint.HTTP.recv_response(conn, ref, 5000)
iex> Mint.HTTP1.close(conn)
iex> response
%{
  status: 200,
  headers: [
    {"date", ...},
    ...
  ],
  body: "{\n  \"user-agent\": \"mint/1.6.2\"\n}\n"
}
```

There is a bunch of libraries which are simply make one off requests, to name a few: tzdata, goth/aws_credentials (they have their own pool, make one off request every ~hour), esbuild/tailwind.

To play the devils advocate, esbuild/tailwind are simply using httpc, copy-pasting these [secure ssl defaults from EEF Security WG](https://erlef.github.io/security-wg/secure_coding_and_deployment_hardening/inets).

### DIFF
--- a/lib/mint/http.ex
+++ b/lib/mint/http.ex
@@ -900,6 +900,11 @@ defmodule Mint.HTTP do
       Contrary to `recv/3`, this function does not return partial responses on errors. Use
       `recv/3` for full control.
 
+  > #### Error {: .error}
+  >
+  > This function can only be used for one-off requests. If there is another concurrent request,
+  > started by `request/5`, it will crash.
+
   ## Options
 
     * `:timeout` - the maximum amount of time in milliseconds waiting to receive the response.
@@ -979,8 +984,9 @@ defmodule Mint.HTTP do
   end
 
   # Ignore entries from other requests.
-  defp recv_response([_entry | rest], acc, conn, ref, timeout) do
-    recv_response(rest, acc, conn, ref, timeout)
+  defp recv_response([entry | _rest], _acc, _conn, _ref, _timeout) when is_tuple(entry) do
+    ref = elem(entry, 1)
+    raise "received unexpected response from request #{inspect(ref)}"
   end
 
   defp recv_response([], acc, conn, ref, timeout) do
--- a/lib/mint/http.ex
+++ b/lib/mint/http.ex
@@ -903,6 +903,9 @@ defmodule Mint.HTTP do
       Contrary to `recv/3`, this function does not return partial responses on errors. Use
       `recv/3` for full control.
 
+  This function only handles responses for the given `request_ref`. Do not use it when making
+  concurrent requests as responses other than for `request_ref` are ignored.
+
   ## Examples
 
       iex> {:ok, conn} = Mint.HTTP.connect(:https, "httpbin.org", 443, mode: :passive)
@@ -922,9 +925,9 @@ defmodule Mint.HTTP do
                headers: [{binary(), binary()}],
                body: binary()
              }
-  def recv_response(conn, ref, timeout)
-      when is_reference(ref) and Util.is_timeout(timeout) do
-    recv_response([], {nil, [], ""}, conn, ref, timeout)
+  def recv_response(conn, request_ref, timeout)
+      when is_reference(request_ref) and Util.is_timeout(timeout) do
+    recv_response([], {nil, [], ""}, conn, request_ref, timeout)
   end
 
   defp recv_response(
--- a/lib/mint/http.ex
+++ b/lib/mint/http.ex
@@ -876,14 +876,11 @@ defmodule Mint.HTTP do
   @version Mix.Project.config()[:version]
 
   @doc """
-  Receives response from the socket in a blocking way.
+  Sends a request and receives a response from the socket in a blocking way.
 
-  This function receives a response for the request identified by `request_ref` on the connection
-  `conn`.
+  This function is a convenience for sending a request with `request/5` and repeatedly calling `recv/3`.
 
-  `timeout` is the maximum time to wait before returning an error.
-
-  This function is a convenience for repeatedly calling `Mint.HTTP.recv/3`. The result is either:
+  The result is either:
 
     * `{:ok, conn, response}` where `conn` is the updated connection and `response` is a map
       with the following keys:
@@ -903,30 +900,48 @@ defmodule Mint.HTTP do
       Contrary to `recv/3`, this function does not return partial responses on errors. Use
       `recv/3` for full control.
 
-  This function only handles responses for the given `request_ref`. Do not use it when making
-  concurrent requests as responses other than for `request_ref` are ignored.
+  ## Options
+
+    * `:timeout` - the maximum amount of time in milliseconds waiting to receive the response.
+      Setting to `:infinity`, disables the timeout. Defaults to `:infinity`.
 
   ## Examples
 
       iex> {:ok, conn} = Mint.HTTP.connect(:https, "httpbin.org", 443, mode: :passive)
-      iex> {:ok, conn, request_ref} = Mint.HTTP.request(conn, "GET", "/user-agent", [], nil)
-      iex> {:ok, _conn, response} = Mint.HTTP.recv_response(conn, request_ref, 5000)
+      iex> {:ok, conn, response} = Mint.HTTP.request_and_response(conn, "GET", "/user-agent", [], nil)
       iex> response
       %{
         status: 200,
         headers: [{"date", ...}, ...],
         body: "{\\n  \\"user-agent\\": \\"#{@version}\\"\\n}\\n"
       }
+      iex> Mint.HTTP.open?(conn)
+      true
   """
-  @spec recv_response(t(), Types.request_ref(), timeout()) ::
-          {:ok, t(), response} | {:error, Types.error()}
+  @spec request_and_response(
+          t(),
+          method :: String.t(),
+          path :: String.t(),
+          Types.headers(),
+          body :: iodata() | nil | :stream,
+          options :: [{:timeout, timeout()}]
+        ) ::
+          {:ok, t(), response}
+          | {:error, t(), Types.error()}
         when response: %{
                status: non_neg_integer(),
                headers: [{binary(), binary()}],
                body: binary()
              }
-  def recv_response(conn, request_ref, timeout)
-      when is_reference(request_ref) and Util.is_timeout(timeout) do
+  def request_and_response(conn, method, path, headers, body, options \\ []) do
+    options = Keyword.validate!(options, timeout: :infinity)
+
+    with {:ok, conn, ref} <- request(conn, method, path, headers, body) do
+      recv_response(conn, ref, options[:timeout])
+    end
+  end
+
+  defp recv_response(conn, request_ref, timeout) when Util.is_timeout(timeout) do
     recv_response([], {nil, [], ""}, conn, request_ref, timeout)
   end
 
--- a/lib/mint/http.ex
+++ b/lib/mint/http.ex
@@ -939,7 +939,7 @@ defmodule Mint.HTTP do
                body: binary()
              }
   def request_and_response(conn, method, path, headers, body, options \\ []) do
-    options = Keyword.validate!(options, timeout: :infinity)
+    options = keyword_validate!(options, timeout: :infinity)
 
     with {:ok, conn, ref} <- request(conn, method, path, headers, body) do
       recv_response(conn, ref, options[:timeout])
@@ -1006,6 +1006,13 @@ defmodule Mint.HTTP do
       {:error, conn, reason, _responses} ->
         {:error, conn, reason}
     end
+  end
+
+  # TODO: Remove when we require Elixir v1.13
+  if Version.match?(System.version(), ">= 1.13.0") do
+    defp keyword_validate!(keyword, values), do: Keyword.validate!(keyword, values)
+  else
+    defp keyword_validate!(keyword, _values), do: keyword
   end
 
   @doc """
--- a/lib/mint/http.ex
+++ b/lib/mint/http.ex
@@ -900,6 +900,9 @@ defmodule Mint.HTTP do
       as well. An error when sending a request **does not** necessarily mean that the connection
       is closed. Use `open?/1` to verify that the connection is open.
 
+      Contrary to `recv/3`, this function does not return partial responses on errors. Use
+      `recv/3` for full control.
+
   ## Examples
 
       iex> {:ok, conn} = Mint.HTTP.connect(:https, "httpbin.org", 443, mode: :passive)
--- a/lib/mint/http.ex
+++ b/lib/mint/http.ex
@@ -965,15 +965,19 @@ defmodule Mint.HTTP do
   defp recv_response([], acc, conn, ref, timeout) do
     start_time = System.monotonic_time(:millisecond)
 
-    with {:ok, conn, entries} <- recv(conn, 0, timeout) do
-      timeout =
-        if is_integer(timeout) do
-          timeout - System.monotonic_time(:millisecond) - start_time
-        else
-          timeout
-        end
+    case recv(conn, 0, timeout) do
+      {:ok, conn, entries} ->
+        timeout =
+          if is_integer(timeout) do
+            timeout - System.monotonic_time(:millisecond) - start_time
+          else
+            timeout
+          end
 
-      recv_response(entries, acc, conn, ref, timeout)
+        recv_response(entries, acc, conn, ref, timeout)
+
+      {:error, conn, reason, _responses} ->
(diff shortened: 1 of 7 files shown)


wojtekmach:
Another idea is to instead have:

```elixir
{:ok, conn} = Mint.HTTP.connect(:https, "httpbin.org", 443)

{:ok, conn, _acc = nil} =
  Mint.HTTP.request(
    conn,
    "GET",
    "/status/user-agent",
    _headers = [],
    _body = nil,
    _timeout = 5000,
    _acc = nil,
    fn conn, entry, acc ->
      IO.inspect(entry)
      {:cont, conn, acc}
    end
  )
```

which does request/recv_response. Yet another idea is to have a single function that does connect/request/recv_response. I think request/8 above is a nice middle ground though.

whatyouhide:
@wojtekmach why would this crash?

```elixir
{:ok, conn} = Mint.HTTP1.connect(:https, "httpbin.org", 443, mode: :passive)
{:ok, conn, ref1} = Mint.HTTP1.request(conn, "GET", "/user-agent", [], nil)
{:ok, conn, ref2} = Mint.HTTP1.request(conn, "GET", "/user-agent", [], nil)
Mint.HTTP1.recv_response(conn, ref1, 5000)
Mint.HTTP1.recv_response(conn, ref2, 5000)
```

`recv_response` could just discard messages that are not for the given ref, no?

`request/8` is a no-go for me, too complex of an API. `recv_response/3` is nice because it separates the APIs, so that you can still use `request/5` and in case then you decide you're done with async stuff and want to receive a response, you have the option to do so in a blocking fashion with `recv_response/3`.

whatyouhide:
By the way, I’m totally in favor for this. 👍

wojtekmach:
Yes I can totally disregard unrelated messages and can document it as such. The only problem is they cannot be recovered. My understanding is in active mode we do selective receive on _conn_, not _request_, and in passive mode we’d have read from the socket. Again I think it’s fine.

Should we support active mode or passive mode or both?

Should we support HTTP/2, ie this becomes a function on Mint.HTTP?

Fair enough about not supporting /8.

whatyouhide:
Yes HTTP/2 for sure. I’m thinking we might want to enforce this on passive conns actually, because if you think about it there's not much advantage to doing a blocking req and using active mode (yeah can recv bytes while parsing but probably negligible).

wojtekmach:
Got it, thanks.

Would you prefer to start with just recv_response/3 or recv_response/5 (that additionally has `acc` and `fun`) or both? In the PR we have both at the moment.

wojtekmach:
I renamed this to `Mint.HTTP.recv_response/3`, added basic docs and tests, and updated PR description accordingly.

wojtekmach:
@whatyouhide interesting! By _stream_ of `{atom, term}` tuples you mean a _list_ right? Or we're talking about lazy streams? I assume it's a list (so this function returns updated `conn`) so a follow-up question, what'd be a benefit over a map? I don't know if this would happen (or even be possible) in practice but maybe a silly server could return `{:data, "foo"}` followed by `{:data, ""}` and this would break passing it to `Map.new()`. A response with trailers would break Map.new too but I concede it's rather rare. If people are going to end up using Enum.reduce over this list a lot of times then I'd push for a version with acc + fun which will be very similar to Enum.reduce. In other words, in that case I'd have `recv_response/3` which returns a map and `recv_response_5` (or `recv_response_while/5`). WDYT?

lib/mint/http.ex
```diff
@@ -872,6 +873,100 @@ defmodule Mint.HTTP do
...
+    recv_response(rest, acc, conn, ref, timeout)
+  end
+
+  defp recv_response([{:headers, ref, headers} | rest], acc, conn, ref, timeout) do
+    acc = update_in(acc.headers, &(&1 ++ headers))
+    recv_response(rest, acc, conn, ref, timeout)
+  end
+
+  defp recv_response([{:data, ref, data} | rest], acc, conn, ref, timeout) do
+    acc = update_in(acc.body, &(&1 <> data))
```
wojtekmach, line 938: This could be:

```suggestion
    acc = update_in(acc.body, &[&1, data])
```

and then on `:done` we IO.iodata_to_binary it. I decided to keep it simple for now, Finch is like that (using `<>`) at the moment too.

josevalim: We shouldn't use `<>` and we should probably change Finch to not do that either. That will most likely copy the left-side on every concatenation. The IO version should be fairly more efficient. I would also keep the `acc` as a triplet.

josevalim (approved):
One nit and looks good to me!

lib/mint/http.ex
wojtekmach: Keep as triplet and convert to map at the end right?

josevalim: Yes!

wojtekmach:
RE CI failure, sigh, https://github.com/actions/runner-images/issues/9692.

wojtekmach:
CI is fixed in #448.

whatyouhide:
Damn I always forget that streams cannot carry an accumulator, and we really want to accumulate over the conn 🙄

wojtekmach:
We could have a single function that does connect/request/recv_response/close and it returns {:ok, enumerable}, no conn, it would be something like request/9 or, you know, request/_1_. :)

whatyouhide:
Since the use case here is to make a request and receive a response in a single operation without having to deal with the intricacies of Mint, then I would go with:

```elixir
Mint.HTTP.request_and_response(
  conn,
  method,
  path,
  headers,
  body,
  options :: [{:timeout, timeout()}]
)
```

which returns a response map (usual `%{status: ..., ...}`). This is a handy compromise for one-shot requests. If you want to have control over accumulation, go with receiving the messages (or passive mode). Thoughts?

wojtekmach:
@whatyouhide yeah, I think this is a good compromise. The only thing that comes to my mind is with a distinct `recv_response/3` we may still support request body streaming before calling it (since it blocks), but with request_and_response/6 we can't. But I think that's probably OK, can always drop down to lower level things. WDYT?

josevalim:
I think I would keep two separate functions. You get to support request streaming, passive mode, and pipelining by keeping them apart. Sure they are not “deal breakers” but we are not gaining much by unifying anyway, so why do it?

wojtekmach:
That being said I think what we have right now is pretty decent (if I may say so)

```elixir
iex> {:ok, conn} = Mint.HTTP.connect(:https, "httpbin.org", 443, mode: :passive)
iex> {:ok, conn, request_ref} = Mint.HTTP.request(conn, "GET", "/user-agent", [], nil)
iex> {:ok, _conn, response} = Mint.HTTP.recv_response(conn, request_ref, 5000)
iex> response
%{
  status: 200,
  headers: [{"date", ...}, ...],
  body: "{\\n  \\"user-agent\\": \\"#{@version}\\"\\n}\\n"
}
```

whatyouhide:
Yeah I like that but I do not like the fact that it can trip you up if there are multiple in-flight reqs. Can we raise an error if there are?

josevalim:
Why can you trip up? You can for HTTP1 if you consume the responses out of order but not for HTTP2, right?

whatyouhide:
You can because as Wojtek said you will still consume the bytes (or msgs) for other requests. Say you have `ref1` and `ref2` in flight and you get these bytes:

```
ref1(HEADERS)
ref2(HEADERS)
ref1(BODY)
ref1(END)
```

in a single packet. If you call `recv_response(ref1)`, then you will consume the headers for `ref2` and won't be able to return them in the response.

wojtekmach:
Yeah see this comment: https://github.com/elixir-mint/mint/pull/447#issuecomment-2258261087. I’m fine with discarding (right now on the branch) or crashing. Lmk!

whatyouhide:
We can crash if there are multiple reqs in flight, the conn knows that.

josevalim:
Sorry, to be clear, there is a valid use case for this feature:

1. HTTP1, as long as you consume them in order
2. HTTP2

So I am not sure why we should raise and forbid all cases? Couldn't we just document it instead that HTTP1 itself requires order?

whatyouhide:
@josevalim okay yes in HTTP/1 reqs are in order so this is not an issue. In HTTP/2, this is an issue, as per https://github.com/elixir-mint/mint/pull/447#issuecomment-2268750627. If frames come out of order, we can absolutely match them back into their corresponding request, but generally Mint emits "responses" as frames. So in the sequence of frames in the comment above, a normal Mint flow would return

```elixir
[{:headers, ref1, ...}, {:headers, ref2, ...}, {:body, ref1, ...}, {:done, ref1}]
```

In this, you didn't lose the headers of `ref2`. If `recv_response` instead returns `%{status, headers, body}` for `ref1`, we lost the headers for `ref2` unless we buffer them into the conn (I would rather not).

Considering that this is a nice utility for one-shot requests, and you can still implement blocking `recv_response` through the normal message-based API anyway, I'd say we don't need HTTP/2 multiplexing?

josevalim:
@whatyouhide If we receive messages from `ref2` when receiving `ref1`, we should store them on the `conn`, and replay them only when you receive `ref2`, no?

whatyouhide:
@josevalim yeah that's the complexity I want to avoid, as I don't think it's justified for the `recv_response` *utility*. We don't store anything now because if you receive a complete frame for `ref2` while `ref1` is also in flight, we just **emit it** (as a `{:status, ...}`, `{:headers, ...}`, etc response).

whatyouhide:
Complexity as we don't need it 🙃

Also, can you describe a use case where I have `ref1` and `ref2` in flight, and then I call `recv_response(ref1)`. I never call `recv_response(ref2)`. What happens to the frames we receive for `ref2`? Do we surface them only when the user finally calls `stream/2`?

whatyouhide:
@josevalim consider that this is just one mor argument than `Mint.HTTP.request/5`, so I think we're okay 🙃 The options will ensure that we won't have any more arguments in the future as we can just add them as options then.

Also, the reason I’m pushing back here on features like body streaming and whatnot is that I want Mint to stay really low level and just an abstraction around the network socket, but with protocol understanding. If we start to add fancy functionality then I’m afraid we will just end up with Finch 😄 A single one-shot req/resp function makes sense for scripting with less deps, and libraries and so on, but I would probably stop there and ask anyone who needs more complex scenarios to just use the normal `stream/2`-based API.

josevalim:
> Also, the reason I’m pushing back here on features like body streaming and whatnot is that I want Mint to stay really low level and just an abstraction around the network socket, but with protocol understanding.

Yes, I agree with this. At the same time, there can be use cases where you only need to do one HTTP request, and people are resorting to Finch because there is nothing slightly more ergonomic in Mint. Also, if the goal is to one day have Mint in Erlang/OTP, I think such conveniences make sense, which is why I am not opposed for some slight exploration here. :) But yeah, if you need body streaming, it is probably fine to rely on the low-level ones.

whatyouhide:
@wojtekmach let's go with `request_and_response` for now then and evolve from there. Body streaming and whatnot might come in later as options after all 🙃

lib/mint/http.ex
```diff
@@ -872,6 +873,148 @@ defmodule Mint.HTTP do
           | {:error, t(), Types.error(), [Types.response()]}
   def recv(conn, byte_count, timeout), do: conn_module(conn).recv(conn, byte_count, timeout)
 
+  @version Mix.Project.config()[:version]
+
+  @doc """
+  Sends a request and receives a response from the socket in a blocking way.
+
+  This function is a convenience for sending a request with `request/5` and repeatedly calling `recv/3`.
```
whatyouhide, line 881:
```suggestion
  This function is a convenience for sending a request with `request/5` and calling `recv/3`
  repeatedly until a full response is received.
```

lib/mint/http.ex
```diff
@@ -872,6 +873,148 @@ defmodule Mint.HTTP do
...
+
+  The result is either:
+
+    * `{:ok, conn, response}` where `conn` is the updated connection and `response` is a map
+      with the following keys:
+
+        * `:status` - HTTP response status, an integer.
+
+        * `:headers` - HTTP response headers, a list of tuples `{header_name, header_value}`.
```
whatyouhide, line 890: Also trailing headers?

lib/mint/http.ex
```diff
@@ -872,6 +873,148 @@ defmodule Mint.HTTP do
...
+
+  > #### Error {: .error}
+  >
+  > This function can only be used for one-off requests. If there is another concurrent request,
+  > started by `request/5`, it will crash.
+
+  ## Options
+
+    * `:timeout` - the maximum amount of time in milliseconds waiting to receive the response.
+      Setting to `:infinity`, disables the timeout. Defaults to `:infinity`.
```
whatyouhide, line 911: Mh, are we ok with a default that potentially blocks forever? Maybe 30s is a better default?

lib/mint/http.ex
```diff
@@ -872,6 +873,148 @@ defmodule Mint.HTTP do
...
+      iex> {:ok, conn} = Mint.HTTP.connect(:https, "httpbin.org", 443, mode: :passive)
+      iex> {:ok, conn, response} = Mint.HTTP.request_and_response(conn, "GET", "/user-agent", [], nil)
+      iex> response
+      %{
+        status: 200,
+        headers: [{"date", ...}, ...],
+        body: "{\\n  \\"user-agent\\": \\"#{@version}\\"\\n}\\n"
+      }
+      iex> Mint.HTTP.open?(conn)
+      true
```
whatyouhide, line 924:
```suggestion
      true

```

Nit, sorry 🙃

lib/mint/http.ex
```diff
@@ -872,6 +873,148 @@ defmodule Mint.HTTP do
...
+          Types.headers(),
+          body :: iodata() | nil | :stream,
+          options :: [{:timeout, timeout()}]
+        ) ::
+          {:ok, t(), response}
+          | {:error, t(), Types.error()}
+        when response: %{
+               status: non_neg_integer(),
+               headers: [{binary(), binary()}],
+               body: binary()
```
whatyouhide, line 939:
```suggestion
               status: Types.status(),
               headers: Types.headers(),
               body: binary()
```

lib/mint/http.ex
```diff
@@ -872,6 +873,148 @@ defmodule Mint.HTTP do
...
+         timeout
+       ) do
+    recv_response(rest, {status, headers ++ new_headers, body}, conn, ref, timeout)
+  end
+
+  defp recv_response([{:data, ref, data} | rest], {status, headers, body}, conn, ref, timeout) do
+    recv_response(rest, {status, headers, [body, data]}, conn, ref, timeout)
+  end
+
+  defp recv_response([{:done, ref} | _rest], {status, headers, body}, conn, ref, _timeout) do
```
whatyouhide, line 977: Shouldn't we error out here if there is trailing response data? That is, this should always be `[{:done, ref}]` no?

lib/mint/http.ex
```diff
@@ -872,6 +873,148 @@ defmodule Mint.HTTP do
...
+
+        recv_response(entries, acc, conn, ref, timeout)
+
+      {:error, conn, reason, _responses} ->
+        {:error, conn, reason}
+    end
+  end
+
+  # TODO: Remove when we require Elixir v1.13
+  if Version.match?(System.version(), ">= 1.13.0") do
```
whatyouhide, line 1012:
```suggestion
  if function_exported?(Keyword, :validate!, 2) do
```

lib/mint/http.ex
```diff
@@ -872,6 +873,148 @@ defmodule Mint.HTTP do
...
+    * `{:error, conn, reason}` where `conn` is the updated connection and `reason` is the cause
+      of the error. It is important to store the returned connection over the old connection in
+      case of errors too, because the state of the connection might change when there are errors
+      as well. An error when sending a request **does not** necessarily mean that the connection
+      is closed. Use `open?/1` to verify that the connection is open.
+
+      Contrary to `recv/3`, this function does not return partial responses on errors. Use
+      `recv/3` for full control.
+
+  > #### Error {: .error}
```
whatyouhide, line 903: We said we'd also raise if the mode is not `:passive`, right?

whatyouhide (commented):
Thanks @wojtekmach this is looking great. Left a few comments 🙃

wojtekmach:
Sorry it's been stale, let me close this for now and I'll re-open when all feedback is addressed.

PR #448, opened 2024-08-03
Update to `docker compose`

wojtekmach:
Docker Compose v1 aka `docker-compose` is deprecated in favour of v2 aka `docker compose`. See https://github.com/actions/runner-images/issues/9692.

Also, set platform linux/amd64 for httpbin. It does not have arm64 image and starting this container fails on arm64 Macs.


whatyouhide:
Yep, thanks @wojtekmach 💟

Issue #453, opened 2024-11-06
Allow not validating target


PragTob:
:wave: 

Hi there everyone and thank you so much for mint - been a joy to use via `req` and helps us make a significant amount of HTTP requests! :green_heart: 

## Problem

In our context we're frequently conftonted with query strings that aren't strictly standard conform. These contain "macros". Those look something like this: `referrer={encSite}` and can occur multiple times per URL.

There are many different patterns of these that I know of, usually occurring in the value part of the query params, some more patterns I know of:

* `key=[%macro%]`
* `key=${macro}`
* `key=%%macro%%`
* `"key=%ebuy!"`

This blows up in `Mint.HTTP1.Request.validate_target!/1` ( and `/2`) https://github.com/elixir-mint/mint/blob/50b11d668b6a240b0d9b20c67fbb41a10a7410b1/lib/mint/http1/request.ex#L49-L65

While I understand enforcing standards, sometimes the real (old) applications out there makes that hard.

<details>
<summary>Small example</summary>

```
iex(1)> HttpService.get!("https://some.url.com/path?rb=51439&ca=20863020&ra=1730132562831&_o=51439&_t=20863020")
# in this example, a redirect happens
** (Req.HTTPError) invalid request target: "/path?c=bd8618c307ae9885a12561b7191e2cea&cid=5134455426882927793&referrer={encSite}&some=more"
    (req 0.5.6) lib/req.ex:1092: Req.request!/2
    iex:1: (file)
```

</details>

## Solution

I'd be really happy about being able to provide an option where validating the target is skipped or made more lenient. Or some way in which I could easily accomplish that. I'm not sure of the consequences of this but yeah :)

(and then get to it through Req and Finch)

I'm happy to implement code to do this, once aligned on a solution.

I also have a bunch of affected URLs in tests (running against bypass) or as examples lying around. So, I can definitely test solutions.

## Workaround

The current workaround we use is that we implemented code to sanitize the query string (working similar to `validate_target!/1`, it's fun as the URLs we get are partially encoded and partially not so running encode on all of it is not what we want).

The topic gets a bit harder though, as sometimes redirects will also include these broken URLs. Which is also what you see in the example I posted above. As a workaround for this we're implementing our own redirect following now (so that we can sanitize the query strings on every hop). This is the part where I decided to open this issue, as I'd really like to get rid of this at least :sweat_smile: (but at best all the custom code I wrote for this).

## Behavior of other HTTP clients

[...]

PR #456, opened 2024-11-08
Allow not validating target

PragTob:
As discussed in https://github.com/elixir-mint/mint/issues/453

I tried to follow the guidance of `case_sensitive_headers` so
that these options are treated somewhat similarly.

I tried to keep with the patterns of surrounding code, if you want anything changed/I missed something - happy to do so!

In a separate commit I also included a small nicety as I saw the definition of the mint user agent duplicated over the tests. Happy to roll that back though.

Some questions/remarks inline.

Thanks for all your work! :green_heart: 

Tasks left for me:
* [ ] Check if this actually fixes what I reported in #453 :)

IMG_20220619_184733

### DIFF
--- a/lib/mint/http1.ex
+++ b/lib/mint/http1.ex
@@ -93,6 +93,7 @@ defmodule Mint.HTTP1 do
     :mode,
     :scheme_as_string,
     :case_sensitive_headers,
+    :skip_target_validation,
     requests: :queue.new(),
     state: :closed,
     buffer: "",
@@ -123,6 +124,10 @@ defmodule Mint.HTTP1 do
     * `:case_sensitive_headers` - (boolean) if set to `true` the case of the supplied
        headers in requests will be preserved. The default is to lowercase the headers
        because HTTP/1.1 header names are case-insensitive. *Available since v1.6.0*.
+    * `:skip_target_validation` - (boolean) if set to `true` the target of a request
+       will not be validated. You might want this if you deal with non standard-
+       conform URIs but need to preserve them. The default is to validate the request
+       target. *Available since v1.?.?*
 
   """
   @spec connect(Types.scheme(), Types.address(), :inet.port_number(), keyword()) ::
@@ -200,7 +205,8 @@ defmodule Mint.HTTP1 do
         scheme_as_string: Atom.to_string(scheme),
         state: :open,
         log: log?,
-        case_sensitive_headers: Keyword.get(opts, :case_sensitive_headers, false)
+        case_sensitive_headers: Keyword.get(opts, :case_sensitive_headers, false),
+        skip_target_validation: Keyword.get(opts, :skip_target_validation, false)
       }
 
       {:ok, conn}
@@ -280,7 +286,8 @@ defmodule Mint.HTTP1 do
              method,
              path,
              Headers.to_raw(headers, conn.case_sensitive_headers),
-             body
+             body,
+             conn.skip_target_validation
            ),
          :ok <- transport.send(socket, iodata) do
       request_ref = make_ref()
--- a/lib/mint/http1.ex
+++ b/lib/mint/http1.ex
@@ -126,8 +126,8 @@ defmodule Mint.HTTP1 do
        because HTTP/1.1 header names are case-insensitive. *Available since v1.6.0*.
     * `:skip_target_validation` - (boolean) if set to `true` the target of a request
        will not be validated. You might want this if you deal with non standard-
-       conform URIs but need to preserve them. The default is to validate the request
-       target. *Available since v1.?.?*
+       conforming URIs but need to preserve them. The default is to validate the request
+       target. *Available since v1.7.0*.
 
   """
   @spec connect(Types.scheme(), Types.address(), :inet.port_number(), keyword()) ::
--- a/lib/mint/http1.ex
+++ b/lib/mint/http1.ex
@@ -281,13 +281,13 @@ defmodule Mint.HTTP1 do
       |> add_default_headers(conn)
 
     with {:ok, headers, encoding} <- add_content_length_or_transfer_encoding(headers, body),
+         :ok <- validate_request_target(path, conn.skip_target_validation),
          {:ok, iodata} <-
            Request.encode(
              method,
              path,
              Headers.to_raw(headers, conn.case_sensitive_headers),
-             body,
-             conn.skip_target_validation
+             body
            ),
          :ok <- transport.send(socket, iodata) do
       request_ref = make_ref()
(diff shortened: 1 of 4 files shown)


lib/mint/http1.ex
```diff
@@ -123,6 +124,10 @@ defmodule Mint.HTTP1 do
     * `:case_sensitive_headers` - (boolean) if set to `true` the case of the supplied
        headers in requests will be preserved. The default is to lowercase the headers
        because HTTP/1.1 header names are case-insensitive. *Available since v1.6.0*.
+    * `:skip_target_validation` - (boolean) if set to `true` the target of a request
+       will not be validated. You might want this if you deal with non standard-
+       conform URIs but need to preserve them. The default is to validate the request
+       target. *Available since v1.?.?*
```
PragTob, line 130: Added it to keep on par with `case_sensititve_headers` but I'm not sure if this would be v 1.7.0 if it was released (technically it's a feature).

Also, should I add it to the Changelog then? :thinking:

lib/mint/http1/request.ex
```diff
@@ -3,9 +3,9 @@ defmodule Mint.HTTP1.Request do
 
   import Mint.HTTP1.Parse
 
-  def encode(method, target, headers, body) do
+  def encode(method, target, headers, body, skip_target_validation \\ false) do
```
PragTob, line 6: this duplicates the default but I didn't:

* want it to be forced to be passed along
* break the existing interface (although technically it's `@moduledoc false`)

Also opted for a plain boolean, could also be turned into an `opts`/kwlist but seemed simpler so far.

test/mint/http1/conn_test.exs
```diff
@@ -760,6 +760,38 @@ defmodule Mint.HTTP1Test do
 
                """)
     end
+
+    @invalid_target "%%"
+    test "targets are validated by default", %{port: port, server_ref: server_ref} do
+      assert {:ok, conn} = HTTP1.connect(:http, "localhost", port)
```
PragTob, line 766: may be overkill as I only added the option to skip, but I wanted the tests to show both cases and I wanted to test on the `conn` level to show that the option is correctly passed on

test/mint/http1/conn_test.exs
```diff
@@ -548,7 +548,7 @@ defmodule Mint.HTTP1Test do
              request_string("""
              GET / HTTP/1.1
              host: localhost:#{port}
-             user-agent: mint/#{Mix.Project.config()[:version]}
+             user-agent: #{mint_user_agent()}
```
PragTob, line 551: This is just a small convenience change I made in a separate commit - hope it's not too much noise but I also didn't wanna do it in a separate PR. Happy to roll it back!

PragTob:
Sorry for the force push, forgot one guiding comment and wanted to have it gone form history :sweat_smile:

PragTob:
Works as expected on the `mint` level:

```
iex(1)> {:ok, conn} = Mint.HTTP.connect(:http, "httpbin.org", 80); {:ok, conn, request_ref} = Mint.HTTP.request(conn, "GET", "?na=%%boom%%", [], "")
** (MatchError) no match of right hand side value: {:error, %Mint.HTTP1{host: "httpbin.org", port: 80, request: nil, streaming_request: nil, socket: #Port<0.5>, transport: Mint.Core.Transport.TCP, mode: :active, scheme_as_string: "http", case_sensitive_headers: false, skip_target_validation: false, requests: {[], []}, state: :open, buffer: "", proxy_headers: [], private: %{}, log: false}, %Mint.HTTPError{reason: {:invalid_request_target, "?na=%%boom%%"}, module: Mint.HTTP1}}
    (stdlib 6.1) erl_eval.erl:652: :erl_eval.expr/6
    iex:1: (file)
iex(1)> {:ok, conn2} = Mint.HTTP.connect(:http, "httpbin.org", 80, skip_target_validation: true); {:ok, conn, request_ref} = Mint.HTTP.request(conn2, "GET", "?na=%%boom%%", [], "")
{:ok,
 %Mint.HTTP1{
   host: "httpbin.org",
   port: 80,
   request: %{
     status: nil,
     version: nil,
     state: :status,
     connection: [],
     body: nil,
     ref: #Reference<0.1432147674.813432838.138316>,
     data_buffer: [],
     headers_buffer: [],
     content_length: nil,
     method: "GET",
     transfer_encoding: []
   },
   streaming_request: nil,
   socket: #Port<0.6>,
   transport: Mint.Core.Transport.TCP,
   mode: :active,
   scheme_as_string: "http",
   case_sensitive_headers: false,
   skip_target_validation: true,
   requests: {[], []},
   state: :open,
   buffer: "",
   proxy_headers: [],
   private: %{},
   log: false
 },
 #Reference<0.1432147674.813432838.138316>}
iex(2)> flush()
{:tcp, #Port<0.6>,
 "HTTP/1.1 400 Bad Request\r\nServer: awselb/2.0\r\nDate: Fri, 08 Nov 2024 11:01:57 GMT\r\nContent-Type: text/html\r\nContent-Length: 122\r\nConnection: close\r\n\r\n<html>\r\n<head><title>400 Bad Request</title></head>\r\n<body>\r\n<center><h1>400 Bad Request</h1></center>\r\n</body>\r\n</html>\r\n"}
:ok
```

Can't figure out how to get it through Req/Finch (I thought I could get them to my Finch pool via `conn_opts` but that seems to not work, but that's not for here to solve :) )

lib/mint/http1.ex
whatyouhide, line 130:
```suggestion
    * `:skip_target_validation` - (boolean) if set to `true` the target of a request
       will not be validated. You might want this if you deal with non standard-
       conforming URIs but need to preserve them. The default is to validate the request
       target. *Available since v1.7.0*.
```

test/mint/http1/conn_test.exs
```diff
@@ -997,4 +1029,6 @@ defmodule Mint.HTTP1Test do
   defp stream_message_bytewise(<<>>, conn, responses) do
     {:ok, conn, responses}
   end
+
+  defp mint_user_agent, do: "mint/#{Mix.Project.config()[:version]}"
```
whatyouhide, line 1033: You can probably do this at compile time in a module attr even

whatyouhide (commented):
Looking already great 💟

lib/mint/http1/request.ex
PragTob: I'll try moving it over, the case sensitive headers option is also handled purely in `Mint.HTTP` - wasn't sure about the separation of concerns. Thanks for the feedback!

PragTob: Moved it over after a short coffee break :coffee: 

One thing that is slightly odd now is that target validation is in `Mint.HTTP1` while header validation is still in `Mint.HTTP1.Request`:

```elixir
  defp encode_headers(headers) do
    Enum.reduce(headers, "", fn {name, value}, acc ->
      validate_header_name!(name)
      validate_header_value!(name, value)
      [acc, name, ": ", value, "\r\n"]
    end)
  end
```

Might be worth to move. The reason I didn't yet is because it'd mean enumerating the headers another time and I'm not sure if that's a performance concern that matters in `mints` context (generally it shouldn't matter for "normal" amounts of headers but yeah).
Also it'd mean me changing more code that I don't know super well :)

Happy to change or keep depending on what you think @whatyouhide !

test/mint/http1/conn_test.exs
PragTob: Works with `[:version]` :)

lib/mint/http1/request.ex
whatyouhide: Let's keep it this way, don't worry about it for now.

whatyouhide:
@PragTob do we need to do this for HTTP/2 too?

PragTob:
@whatyouhide I'm unsure :) I'm guessing so but I didn't find any similar code with HTTP2 on a cursory look and I'm also not as familiar with HTTP2, let me see if I can find something.

PragTob:
@whatyouhide as best as I can see there is no validation going on in the HTTP2 implementation.

`path` is taken here: https://github.com/elixir-mint/mint/blob/8516bcda190a644a5fe960b13869107ff76ee9b3/lib/mint/http2.ex#L510-L527

and only passed on to this function where it is just added to the header:

https://github.com/elixir-mint/mint/blob/8516bcda190a644a5fe960b13869107ff76ee9b3/lib/mint/http2.ex#L1354-L1370

there are some further validations going on, but as far as I can see none of them are for the path in particular/for a similar problem.

whatyouhide:
Coolsies, then we're good 🙃

PR #460, opened 2025-01-24
Fix warnings on Elixir v1.19

josevalim:


josevalim:
Because the macro doesn't work :)

whatyouhide:
@josevalim could you elaborate on why?

josevalim:
I just doesn't work, converting into a macro would be this:

```
case data do
  %Foo{} -> Foo
  %Bar{} -> Bar
  %Baz{} -> Baz
end.something(data)
```

So the type system will still see you are trying to invoke Foo with something that could be %Baz{}. We would need a macro that expanded to this:

```
case data do
  %Foo{} -> Foo.something(data)
  %Bar{} -> Bar.something(data)
  %Baz{} -> Baz.something(data)
end
```

Which is basically the same as conn_apply anyway.

whatyouhide:
Aaah gotcha ok. Thanks 🙃

PR #469, opened 2025-08-07
Forbid or replace empty targets in HTTP/1.1 requests

ruslandoga:
👋 

Right now Mint allows sending empty target in the request line for HTTP/1.1. I think it should either forbid such requests or [replace](https://datatracker.ietf.org/doc/html/rfc7230#section-5.3.1) them with `/`:

> If the target URI's path component is
   empty, the client MUST send "/" as the path within the origin-form of
   request-target.

### Current behaviour

```elixir
Mint.HTTP1.request(conn, "POST", _path = "", headers, body)
```

this results in `POST  HTTP/1.1`

475595741-7dd92fd2-9c97-424a-ad20-ece2c075e345

### This PR

Forbidding seems easier since replacing `""` with `"/"` might be surprising and not always correct, and there is already a `:skip_target_validation` option to disable "forbidding" if necessary.

### DIFF
--- a/lib/mint/http1.ex
+++ b/lib/mint/http1.ex
@@ -978,6 +978,7 @@ defmodule Mint.HTTP1 do
   # Percent-encoding is not case sensitive so we have to account for lowercase and uppercase.
   @hex_characters ~c"0123456789abcdefABCDEF"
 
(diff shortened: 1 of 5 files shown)


whatyouhide:
@ruslandoga can we add a test that verifies the behavior of this with `:skip_target_validation`?

whatyouhide (approved):

PR #474, opened 2026-01-15
Fix warnings in Elixir 1.20 and require Elixir 1.15

sabiwara:


sabiwara:
@whatyouhide sorry I only ran them on 1.19.
Will try to revert the `require` one for now to see if it fixes it.

sabiwara:
Hum actually even `main` fails on 1.15, seems unrelated to the changes 🤔

whatyouhide:
Thank you @sabiwara 🫶

PR #477, opened 2026-03-19

tank-bohr:
Hey guys. Let me explain the problem I faced with. Below is the typical code for the sreaming body request

```elixir
defmodule MintDialyzerRepro do
  @default_chunk_size 16_384

  def run(method, %URI{scheme: scheme} = uri, headers, body, opts)
      when scheme in ~w[http https] do
    scheme = String.to_existing_atom(scheme)
    port = uri.port || URI.default_port(uri.scheme)
    path = URI.to_string(%URI{uri | scheme: nil, host: nil, authority: nil})

    with {:ok, conn} <- Mint.HTTP.connect(scheme, uri.host, port, opts) do
      stream(conn, method, path, headers, body)
    end
  end

  defp stream(conn, method, path, headers, body) do
    with {:ok, conn, ref} <- Mint.HTTP.request(conn, method, path, headers, :stream) do
      stream_body(conn, ref, body)
    end
  end

  defp stream_body(conn, ref, "") do
    Mint.HTTP.stream_request_body(conn, ref, :eof)
  end

  defp stream_body(conn, ref, body) do
    chunk_size = min(chunk_size(conn, ref), byte_size(body))
    <<chunk::binary-size(chunk_size), rest::binary>> = body

    with {:ok, conn} <- Mint.HTTP.stream_request_body(conn, ref, chunk) do
      stream_body(conn, ref, rest)
    end
  end

  defp chunk_size(conn, ref) do
    case Mint.HTTP.protocol(conn) do
      :http1 ->
        @default_chunk_size

      :http2 ->
        chunk_size_http2(conn, ref)
    end
  end

  defp chunk_size_http2(conn, ref) do
    min(
      Mint.HTTP2.get_window_size(conn, :connection),
      Mint.HTTP2.get_window_size(conn, {:request, ref})
    )
  end
end
```

This code generates dialyzer warnings

```
Total errors: 2, Skipped: 0, Unnecessary Skips: 0
done in 0m1.78s
lib/mint_dialyzer_repro.ex:46:34:call_without_opaque
Function call without opaqueness type mismatch.

Call does not have expected opaque term of type Mint.HTTP2.t() in the 1st position.

Mint.HTTP2.get_window_size(_conn :: Mint.HTTP.t(), :connection)

________________________________________________________________________________
lib/mint_dialyzer_repro.ex:47:34:call_without_opaque
Function call without opaqueness type mismatch.

Call does not have expected opaque term of type Mint.HTTP2.t() in the 1st position.

Mint.HTTP2.get_window_size(_conn :: Mint.HTTP.t(), {:request, reference()})

________________________________________________________________________________
done (warnings were emitted)
Halting VM with exit status 2
```

It was discussed in the https://github.com/elixir-mint/mint/issues/380 and was left with a dirty workaround.
And people are [using](https://github.com/coryodaniel/k8s/blob/590da377bf4783d231f8be30fdcd36434f0a6dce/lib/k8s/client/mint/request.ex#L212-L225) the workaround. And on the OTP-28 such trick in the k8s library leads to the infinite hanging during the PLT building phase.

But it's a legit issue. And it's totally worth to be fixed.

## Opaqueness

Since the caller have to know the exact type for streaming, the `Mint.HTTP.t()` **cannot be opaque**. And the caller cannot know the exact type in advance either, because it's not their decision, but the server's decision.

## Missing API for streaming

Another consequence  of this code and dialyzer warning is the request body streaming API is not sufficient for using it without knowing the underlying protocol. But it's kinda designed to be such. To close the gap the current PR is introducing the missing piece - an ability to get the chunk size for the streaming.

### DIFF
--- a/lib/mint/http.ex
+++ b/lib/mint/http.ex
@@ -126,7 +126,7 @@ defmodule Mint.HTTP do
 
   @behaviour Mint.Core.Conn
 
-  @type t() :: Mint.HTTP1.t() | Mint.HTTP2.t()
+  @type t() :: Mint.HTTP1.t() | Mint.HTTP2.t() | Mint.UnsafeProxy.t()
 
   defguardp is_data_message(message)
             when elem(message, 0) in [:ssl, :tcp] and tuple_size(message) == 3
--- a/lib/mint/http.ex
+++ b/lib/mint/http.ex
@@ -126,7 +126,7 @@ defmodule Mint.HTTP do
 
   @behaviour Mint.Core.Conn
 
-  @opaque t() :: Mint.HTTP1.t() | Mint.HTTP2.t()
+  @type t() :: Mint.HTTP1.t() | Mint.HTTP2.t()
 
   defguardp is_data_message(message)
             when elem(message, 0) in [:ssl, :tcp] and tuple_size(message) == 3
@@ -1064,6 +1064,12 @@ defmodule Mint.HTTP do
   @doc false
   @impl true
   def put_proxy_headers(conn, headers), do: conn_apply(conn, :put_proxy_headers, [conn, headers])
+
+  @impl true
+  def get_send_window(conn, ref), do: conn_apply(conn, :get_send_window, [conn, ref])
+
+  @spec next_body_chunk_size(t(), term(), binary()) :: non_neg_integer()
+  def next_body_chunk_size(conn, ref, body), do: min(get_send_window(conn, ref), byte_size(body))
 
   ## Helpers
 
--- a/lib/mint/http1.ex
+++ b/lib/mint/http1.ex
@@ -102,7 +102,8 @@ defmodule Mint.HTTP1 do
     proxy_headers: [],
     private: %{},
     log: false,
-    optional_responses: []
+    optional_responses: [],
+    sndbuf: nil
   ]
 
   defmacrop log(conn, level, message) do
@@ -204,6 +205,7 @@ defmodule Mint.HTTP1 do
     end
 
     with :ok <- Util.inet_opts(transport, socket),
+         {:ok, sndbuf_opts} <- transport.getopts(socket, [:sndbuf]),
          :ok <- if(mode == :active, do: transport.setopts(socket, active: :once), else: :ok) do
       conn = %__MODULE__{
         transport: transport,
@@ -216,7 +218,8 @@ defmodule Mint.HTTP1 do
         log: log?,
         case_sensitive_headers: Keyword.get(opts, :case_sensitive_headers, false),
         skip_target_validation: Keyword.get(opts, :skip_target_validation, false),
-        optional_responses: validate_optional_response_values(opts)
+        optional_responses: validate_optional_response_values(opts),
+        sndbuf: sndbuf_opts[:sndbuf]
       }
 
       {:ok, conn}
@@ -666,6 +669,10 @@ defmodule Mint.HTTP1 do
   def put_proxy_headers(%__MODULE__{} = conn, headers) when is_list(headers) do
     %{conn | proxy_headers: headers}
   end
+
+  @impl true
+  def get_send_window(%__MODULE__{streaming_request: %{ref: ref}, sndbuf: sndbuf}, ref),
+    do: sndbuf
 
   ## Helpers
 
--- a/lib/mint/http2.ex
+++ b/lib/mint/http2.ex
@@ -1039,6 +1039,10 @@ defmodule Mint.HTTP2 do
     %{conn | proxy_headers: headers}
   end
 
+  @impl true
+  def get_send_window(%__MODULE__{} = conn, ref),
+    do: min(get_window_size(conn, :connection), get_window_size(conn, {:request, ref}))
+
   ## Helpers
 
   defp handle_closed(conn) do
--- a/lib/mint/core/conn.ex
+++ b/lib/mint/core/conn.ex
@@ -62,4 +62,6 @@ defmodule Mint.Core.Conn do
   @callback put_proxy_headers(conn(), Mint.Types.headers()) :: conn()
 
   @callback put_log(conn(), boolean()) :: conn()
+
+  @callback get_send_window(conn(), term()) :: non_neg_integer()
 end
--- a/lib/mint/unsafe_proxy.ex
+++ b/lib/mint/unsafe_proxy.ex
@@ -199,4 +199,9 @@ defmodule Mint.UnsafeProxy do
   def put_proxy_headers(%__MODULE__{}, _headers) do
     raise "invalid function for proxy unsafe proxy connections"
   end
+
+  @impl true
+  def get_send_window(%__MODULE__{module: module, state: state}, ref) do
+    module.get_send_window(state, ref)
+  end
 end
--- a/test/http_test.exs
+++ b/test/http_test.exs
@@ -1,4 +1,44 @@
 defmodule Mint.HTTPTest do
   use ExUnit.Case, async: true
   doctest Mint.HTTP
+
+  alias Mint.{HTTP, HTTP1.TestServer}
+
+  setup do
+    {:ok, port, server_ref} = TestServer.start()
+    assert {:ok, conn} = HTTP.connect(:http, "localhost", port)
+    assert_receive {^server_ref, server_socket}
+
+    [conn: conn, server_socket: server_socket]
+  end
+
+  describe "get_send_window/2" do
+    test "returns a positive integer for an active streaming request", %{conn: conn} do
+      {:ok, conn, ref} = HTTP.request(conn, "GET", "/", [], :stream)
+
+      send_window = HTTP.get_send_window(conn, ref)
+      assert is_integer(send_window)
+      assert send_window > 0
+    end
+  end
+
+  describe "next_body_chunk_size/3" do
+    test "returns body size when body is smaller than send window", %{conn: conn} do
+      {:ok, conn, ref} = HTTP.request(conn, "GET", "/", [], :stream)
+      small_body = "hello"
+      assert HTTP.next_body_chunk_size(conn, ref, small_body) == byte_size(small_body)
+    end
+
+    test "returns send window when body is larger than send window", %{conn: conn} do
+      {:ok, conn, ref} = HTTP.request(conn, "GET", "/", [], :stream)
+      send_window = HTTP.get_send_window(conn, ref)
+      large_body = :binary.copy(<<0>>, send_window + 1000)
+      assert HTTP.next_body_chunk_size(conn, ref, large_body) == send_window
+    end
+
+    test "returns 0 for empty body", %{conn: conn} do
+      {:ok, conn, ref} = HTTP.request(conn, "GET", "/", [], :stream)
+      assert HTTP.next_body_chunk_size(conn, ref, "") == 0
+    end
+  end
 end
--- a/test/mint/http1/conn_test.exs
+++ b/test/mint/http1/conn_test.exs
@@ -1073,6 +1073,21 @@ defmodule Mint.HTTP1Test do
     {:ok, conn, responses}
   end
 
+  describe "get_send_window/2" do
+    test "returns a positive integer for an active streaming request", %{conn: conn} do
+      {:ok, conn, ref} = HTTP1.request(conn, "GET", "/", [], :stream)
+
+      send_window = HTTP1.get_send_window(conn, ref)
+      assert is_integer(send_window)
+      assert send_window > 0
+    end
+
+    test "returns the cached sndbuf value", %{conn: conn} do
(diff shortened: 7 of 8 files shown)


tank-bohr:
@ericmj Could you please take a look

whatyouhide:
What if you make

```elixir
@type Mint.HTTP.t() :: Mint.HTTP1.t() | Mint.HTTP2.t()
```

so that the "top-level" type is not opaque, but the inner types are?

tank-bohr:
> What if you make
> 
> ```elixir
> @type Mint.HTTP.t() :: Mint.HTTP1.t() | Mint.HTTP2.t()
> ```
> 
> so that the "top-level" type is not opaque, but the inner types are?

@whatyouhide That's exactly what my PR does. It un-opaques (exposes) the Mint.HTTP.t(). This will help to solve the dialyzer issue in the existing code (like k8s). Because existing code has to know about internals.

And `get_send_window` callback and `Mint.HTTP.next_body_chunk_size/3` potentially can allow to opaque the `Mint.HTTP.t()` type as well. But it requires changing the callers code. But it's helpful anyhow. It brings the missing piece to the ubiquities API of `Mint.HTTP`

tank-bohr:
the original example with the get_send_window will look like

```elixir
defmodule MintDialyzerRepro do
  def run(method, %URI{scheme: scheme} = uri, headers, body, opts)
      when scheme in ~w[http https] do
    scheme = String.to_existing_atom(scheme)
    port = uri.port || URI.default_port(uri.scheme)
    path = URI.to_string(%URI{uri | scheme: nil, host: nil, authority: nil})

    with {:ok, conn} <- Mint.HTTP.connect(scheme, uri.host, port, opts) do
      stream(conn, method, path, headers, body)
    end
  end

  defp stream(conn, method, path, headers, body) do
    with {:ok, conn, ref} <- Mint.HTTP.request(conn, method, path, headers, :stream) do
      stream_body(conn, ref, body)
    end
  end

  defp stream_body(conn, ref, "") do
    Mint.HTTP.stream_request_body(conn, ref, :eof)
  end

  defp stream_body(conn, ref, body) do
    chunk_size = Mint.HTTP.next_body_chunk_size(conn, ref, body)
    <<chunk::binary-size(chunk_size), rest::binary>> = body

    with {:ok, conn} <- Mint.HTTP.stream_request_body(conn, ref, chunk) do
      stream_body(conn, ref, rest)
    end
  end
end
```

As you see no need for using the HTTP2 module directly. No need to detect protocol. The Mint.HTTP module dispatches to the correct implementation itself. With this change we can keep `Mint.HTTP.t()` opaque. But I still recommend to unopaque it because of the existing code

lib/mint/http1.ex
```diff
@@ -667,6 +670,10 @@ defmodule Mint.HTTP1 do
     %{conn | proxy_headers: headers}
   end
 
+  @impl true
+  def get_send_window(%__MODULE__{streaming_request: %{ref: ref}, sndbuf: sndbuf}, ref),
```
ericmj, line 674: This is semantically very different to what what the HTTP2 module does. There is no concept of send windows in HTTP/1 so we should not expose it on this module.

lib/mint/http.ex
```diff
@@ -126,7 +126,7 @@ defmodule Mint.HTTP do
 
   @behaviour Mint.Core.Conn
 
-  @opaque t() :: Mint.HTTP1.t() | Mint.HTTP2.t()
+  @type t() :: Mint.HTTP1.t() | Mint.HTTP2.t()
```
```suggestion
  @type t() :: Mint.HTTP1.t() | Mint.HTTP2.t() | Mint.UnsafeProxy.t()
```

lib/mint/http2.ex
```diff
@@ -1039,6 +1039,10 @@ defmodule Mint.HTTP2 do
     %{conn | proxy_headers: headers}
   end
 
+  @impl true
+  def get_send_window(%__MODULE__{} = conn, ref),
```
ericmj, line 1043: `get_window_size/2` is already public, I don't think we need a function that only calls `min` on top of that.

lib/mint/http.ex
```diff
@@ -1065,6 +1065,12 @@ defmodule Mint.HTTP do
   @impl true
   def put_proxy_headers(conn, headers), do: conn_apply(conn, :put_proxy_headers, [conn, headers])
 
+  @impl true
+  def get_send_window(conn, ref), do: conn_apply(conn, :get_send_window, [conn, ref])
+
+  @spec next_body_chunk_size(t(), term(), binary()) :: non_neg_integer()
+  def next_body_chunk_size(conn, ref, body), do: min(get_send_window(conn, ref), byte_size(body))
```
ericmj, line 1072: I think this function is unnecessary.

lib/mint/http2.ex
tank-bohr: If it's simple it doesn't mean it's useless. If the vast majority of the code that uses streaming is written this way IMHO it makes sense to implement it in the library. Like a direction, like a guideline. Like it's most likely what you need. And if you need something special use the lower level API which is also available. And people need chunk size for streaming most of the time

lib/mint/http1.ex
tank-bohr: I'm open for suggestions about naming. It was hard to come up with a good name which will reflect the meaning of the concept for both http1 and http2. But streaming exists, and it works for both http1 and http2. There is a number that plays the role of the http2 send window in the http1 streaming. Usually the arbitrary number is used. But I thought that the good default could be the biggest number possible in the current TCP configuration. Conceptually it's the same as HTTP2 does, it's just provided by the underlying protocol

ericmj: sndbuf serves a different purpose to H2 window size and has different behavior if you exceed it, blocking vs erroring. sndbuf can be relevant for H2 also, the TCP send buffer can be smaller than the window size, but I don't know how likely that is to happen.

If the purpose is to ensure you never block in gen_tcp.send then you should probably check sndbuf for H2 also or set send_timeout on the socket.

If we don't currently support these options then please send a separate PR.

lib/mint/http2.ex
ericmj: I would rather add a comment that describes the min pattern or find a better name for get_send_window, it doesn't match the API for get_window_size:

* Why does one have have _size in the name and not the other?
* Why does one have "send" in the name but not the other, they are both send windows?
* Just from the names I can't say what the difference is between get_send_window and get_window_size, I would have to read the docs.

I can't come up with a really good name, maybe "possible_send_size"?

tank-bohr: Name `get_send_window` is awful, I agree. `possible_send_size` sounds better to me. Maybe `possible_chunk_size` or `streaming_chunk_size`. Idk either

lib/mint/http1.ex
tank-bohr: sndbuf in this context is not related to h2 at all. The main goal is to unify the streaming API. For streaming you need the chunk size. For http1.1 you need to get this number somehow. `sndbuf` is just the biggest reasonable chunk size for http1.1 streaming. At least that's how I understand the protocol. Naming is bad and to be discussed. But the idea of giving a universal API for streaming is still valid I believe

ericmj: If sndbuf is required for streaming H1 it is also required for H2. sndbuf is part of gen_tcp and both H1 and H2 use TCP.

The H2 send window is a contract with the receiving server we must obey, which is why Mint errors if you exceed it. sndbuf on the other hand is the size of the senders TCP send buffer, it only relates to the sender, it's not a contract with anyone. You would want to check it or tweak it for performance reasons.

Send windows as a concept does not exist for H1 so it's impossible to unify the APIs. The function should only exist on the HTTP2 module.

tank-bohr: > If sndbuf is required for streaming H1 it is also required for H2. sndbuf is part of gen_tcp and both H1 and H2 use TCP.

It's not required. It's just a number. it's part of TCP protocol, yes

> The H2 send window is a contract with the receiving server we must obey, which is why Mint errors if you exceed it. sndbuf on the other hand is the size of the senders TCP send buffer, it only relates to the sender, it's not a contract with anyone. You would want to check it or tweak it for performance reasons.

I don't propose the contract. I propose a helper for the streaming users. To get the best possible chunk size for streaming (you can think of it as for performance reasons)

> Send windows as a concept does not exist for H1 so it's impossible to unify the APIs. The function should only exist on the HTTP2 module.

Yes. And it exists only for the HTTP2. But the streaming is a common concept. H2 uses send windows for streaming. And http1 doesn't have it. So the best possible (the closest) alternative and concept for the chunk size we can offer to the API consumer is a sndbuf.

tank-bohr:
@ericmj I have created a small PR #483 to fix a dialyzer issue. Because it can bring an immediate value. And we can proceed the discussion around the streaming API within this PR. I encourage you to take a closer look at the code sample in the PR description. It demonstrates the gap in the streaming API I want to cover.

- send window is an H2 concept
- http1 doesn't have it
- h2 uses window to calculate chunk size for streaming
- http1 also supports streaming
- http1 as a chunk size can use either arbitrary size or the biggest possible size on the current TCP configuration
- I believe the second option is better

lib/mint/http1.ex
ericmj: > So the best possible (the closest) alternative and concept for the chunk size we can offer to the API consumer is a sndbuf.

I don't agree with that, it's a different concept.

ericmj:
I believe https://github.com/elixir-mint/mint/pull/483 solves the streaming issue for H2 as you can match on the connection type and call `Mint.HTTP2` directly for the window size. H1 should already work for streaming.

If you need sndbuf for performance reasons you can access it via:
```elixir
{:ok, conn} = Mint.HTTP.connect(:http, host, port, transport_opts: [sndbuf: 1_048_576])

socket = Mint.HTTP.get_socket(conn)
:inet.getopts(socket, [:sndbuf])
# or
:ssl.getopts(socket, [:sndbuf])

tank-bohr:
@ericmj You can forget about sndbuf. It's not important. Do you agree that streaming API is incomplete? We need a chunk size and the current API doesn't give a protocol-agnostic way to get the chunk size

ericmj:
You can't make it protocol agnostic because for H1 the chunk size is arbitrary and for H2 it's <window_size.

The only way I can think that is protocol agnostic would be `send_as_much_as_possible(conn, body) :: binary()` that for H2 splits the binary on window size, sends what fit inside the window and returns what was outside the window, and for H1 always just sends everything and returns an empty binary.

Mint is not supposed to be fully protocol agnostic. It's a low level library that abstracts away the protocol version when it's possible without adding abstraction leaks.

tank-bohr:
> You can't make it protocol agnostic because for H1 the chunk size is arbitrary and for H2 it's <window_size.

Correct. But we can give user a hand and cover the most obvious and straightforward usecase

> send_as_much_as_possible

for h1 it effectively means "send sndbuf bytes"

> Mint is not supposed to be fully protocol agnostic. It's a low level library that abstracts away the protocol version when it's possible without adding abstraction leaks.

Yes! And my proposal is to consider if it's possible to make the streaming protocol-agnostic. Because I believe it's possible. Chunk size is a very low-level thingy. It keep all the flexibility of current API and just helps users not to write the exact same code over and over again

ericmj:
> for h1 it effectively means "send sndbuf bytes"

No, that's only if you don't want to block on the `send` function. And even then there is no guarantee you won't block because you don't know if the receiver will have ACKed before you call `send` again. And there is no guarantee that H2 won't block on `send` for the same reason.

tank-bohr:
> No, that's only if you don't want to block on the send function. 

But why would you want to block? It's easy to make non-blocking call blocking but not other way around. I think that streaming by default should be non-blocking. And I would consider the non-blocking scenario is the default one on the API level. Especially taking into account that it's the only way to send body. A non-blocking call gives a user all the control

tank-bohr:
> And even then there is no guarantee you won't block because you don't know if the receiver will have ACKed before you call send again. And there is no guarantee that H2 won't block on send for the same reason.

I can accept it. But I would expect from underlying library to make the best effort to stream even without guarantees

ericmj:
> But why would you want to block? It's easy to make non-blocking call blocking but not other way around. I think that streaming by default should be non-blocking 

If you want to make it non blocking you need to use something like `send_timeout` at the transport level. The kernel doesn't really expose the available send buffer size so you can't chunk by it. https://www.erlang.org/doc/apps/kernel/gen_tcp.html#non_blocking_send

> I can accept it. But I would expect from underlying library to make the best effort to stream even without guarantees

What we have right now is the best effort, even if using `sndbuf` there is nothing stopping you from making two chunks at the `sndbuf` size and sending them immediately after each other and the second send will likely block.

ericmj:
Thank you for https://github.com/elixir-mint/mint/pull/483. I am closing this issue, if there are things we don't expose at the transport level to make non-blocking easier please open a separate issue or PR. Thanks! 💜

PR #478, opened 2026-04-11
Support Elixir ~> 1.12

ericmj:
Replace pinned variables in binary pattern sizes (`^var`) with `erlang.split_binary/2` to support both Elixir 1.12+ and 1.20+ without warnings.

Need this to vendor Mint into Hex.


whatyouhide (approved):
Great!

PR #480, opened 2026-04-12
Add Mint.HTTP2.set_window_size/3

ericmj:
Advertises a larger HTTP/2 receive window (connection-level or per-stream) by sending a WINDOW_UPDATE frame. Needed because RFC 7540 makes the connection-level initial window tunable only via WINDOW_UPDATE — not SETTINGS — leaving the spec default of 64 KB as the only reachable value without an API like this.

In hex's `mix deps.get` — many parallel multi-MB tarball downloads sharing one HTTP/2 connection — raising the connection window from 64 KB to 8 MB via this function drops 10 runs from 32.7s to 29.2s (10.8%). The time is for the whole `mix deps.get` command so the actual speed up for the HTTP requests is actually a higher percentage.

Deliberately asymmetric with get_window_size/2 (which returns the client *send* window). Docstrings on both carry warning callouts spelling out send-vs-receive so callers don't assume they round-trip.

Target is :connection or {:request, ref}; grow-only (shrink attempts return {:error, conn, %HTTPError{reason: :window_size_too_small}}); new_size validated against 1..2^31-1. Tracks the advertised peak on new receive_window_size fields on the connection and stream.

### DIFF
--- a/lib/mint/http2.ex
+++ b/lib/mint/http2.ex
@@ -1988,9 +1988,14 @@ defmodule Mint.HTTP2 do
     end
   end
 
+  valid_client_settings_without_iws = @valid_client_settings -- [:initial_window_size]
+
   defp apply_client_settings(conn, client_settings) do
     Enum.reduce(client_settings, conn, fn
-      {setting, value}, conn when setting in @valid_client_settings ->
+      {:initial_window_size, initial_window_size}, conn ->
+        update_client_initial_window_size(conn, initial_window_size)
+
+      {setting, value}, conn when setting in unquote(valid_client_settings_without_iws) ->
         update_in(conn.client_settings, &%{&1 | setting => value})
 
       {setting, _value}, _conn ->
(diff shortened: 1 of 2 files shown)


lib/mint/http2.ex
```diff
@@ -797,6 +818,119 @@ defmodule Mint.HTTP2 do
     end
   end
 
+  @doc """
+  Advertises a larger client **receive** window to the server.
```
whatyouhide, line 822: @ericmj are we actually advertising a "larger" window size, or is this instead just letting you set whatever receive window you want (larger *or* smaller)?

ericmj: With the current implementation we can only advertise larger window sizes as `WINDOW_UPDATE` is increment only. We could rework it to be able to set any size (including smaller) by tracking an advertised window size and target window size and not send `WINDOW_UPDATE` until we hit the target, this would slightly increase the complexity of the implementation.

If we don't want to do the rework we can rename the function to `increase_window_size` to make it clearer that it's only larger windows.

What do you think?
1. Keep as-is
2. Rework to set any window size
3. Rename to `increase_window_size`

ericmj: I am leaning towards 2.

whatyouhide: @ericmj what would go wrong if we just send `WINDOW_UPDATE`s with the window size provided by the user?

ericmj: The window update is relative and we cannot give negative numbers so we cannot shrink that way

whatyouhide: Aaaaha then this name is good 👍

whatyouhide:
@ericmj I pushed a commit suggested by Codex—I think it makes sense as we do the same thing for the server side. Thoughts?

ericmj:
@whatyouhide Thanks, that's a nice catch.

Are we good to go on this? Can I get your ✅ ?

EDIT: And could you take a look at https://github.com/elixir-mint/mint/pull/481 🙏

PR #482, opened 2026-04-16
Run HTTP/2 integration tests against local Caddy instead of nghttp2.org

ericmj:
The tests hitting nghttp2.org/httpbin/ had been failing on every CI run because the upstream httpbin backend behind nghttpx returns 502 to GitHub Actions runners. Enable h2c on the Caddy docker service (via a new HTTPBIN_H2C_PORT, default 8081, since the existing HTTPBIN_HTTP_PORT mapping goes directly to httpbin), point the TCP h2c test at it, and drop the duplicate "nghttp2.org/httpbin" describe block (ping and GET are already covered by the local "httpbin.org" block).

### DIFF
--- a/test/mint/http2/integration_test.exs
+++ b/test/mint/http2/integration_test.exs
@@ -32,14 +32,10 @@ defmodule HTTP2.IntegrationTest do
     end
   end
 
-  test "TCP - nghttp2.org" do
-    assert {:ok, %HTTP2{} = conn} = HTTP2.connect(:http, "nghttp2.org", 80)
+  test "TCP - h2c prior knowledge" do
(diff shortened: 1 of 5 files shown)


test/mint/http2/integration_test.exs
```diff
@@ -159,32 +155,6 @@ defmodule HTTP2.IntegrationTest do
     end
   end
 
-  describe "nghttp2.org/httpbin" do
```
ericmj, line 162: These tests duplicated mint code paths already covered by the "httpbin.org" describe block (local Caddy over HTTPS/h2): HTTP2.ping/1 and HTTP2.request/5.

PR #483, opened 2026-04-25
Change Mint.HTTP.t() from opaque to open type

tank-bohr:
as it causes the infinite loop hanging of dialyzer on OTP-26

See also discussion in #477


ericmj:
Thank you! 💜

PR #484, opened 2026-04-26

tank-bohr:
Continues the discussion from the #477

Before

```elixir
defmodule MintHttpStreaming do
  @default_chunk_size 16_384

  def run(method, %URI{scheme: scheme} = uri, headers, body, opts)
      when scheme in ~w[http https] do
    scheme = String.to_existing_atom(scheme)
    port = uri.port || URI.default_port(uri.scheme)
    path = URI.to_string(%URI{uri | scheme: nil, host: nil, authority: nil})

    with {:ok, conn} <- Mint.HTTP.connect(scheme, uri.host, port, opts) do
      stream(conn, method, path, headers, body)
    end
  end

  defp stream(conn, method, path, headers, body) do
    with {:ok, conn, ref} <- Mint.HTTP.request(conn, method, path, headers, :stream) do
      stream_body(conn, ref, body)
    end
  end

  defp stream_body(conn, ref, "") do
    Mint.HTTP.stream_request_body(conn, ref, :eof)
  end

  defp stream_body(conn, ref, body) do
    chunk_size = min(chunk_size(conn, ref), byte_size(body))
    <<chunk::binary-size(chunk_size), rest::binary>> = body

    with {:ok, conn} <- Mint.HTTP.stream_request_body(conn, ref, chunk) do
      stream_body(conn, ref, rest)
    end
  end

  defp chunk_size(conn, ref) do
    case Mint.HTTP.protocol(conn) do
      :http1 ->
        @default_chunk_size

      :http2 ->
        chunk_size_http2(conn, ref)
    end
  end

  defp chunk_size_http2(conn, ref) do
    min(
      Mint.HTTP2.get_window_size(conn, :connection),
      Mint.HTTP2.get_window_size(conn, {:request, ref})
    )
  end
end
```

After

```elixir
defmodule MintHttpStreaming do
  def run(method, %URI{scheme: scheme} = uri, headers, body, opts)
      when scheme in ~w[http https] do
    scheme = String.to_existing_atom(scheme)
    port = uri.port || URI.default_port(uri.scheme)
    path = URI.to_string(%URI{uri | scheme: nil, host: nil, authority: nil})

    with {:ok, conn} <- Mint.HTTP.connect(scheme, uri.host, port, opts) do
      stream(conn, method, path, headers, body)
    end
  end

  defp stream(conn, method, path, headers, body) do
    with {:ok, conn, ref} <- Mint.HTTP.request(conn, method, path, headers, :stream) do
      stream_body(conn, ref, body)
    end
  end

  defp stream_body(conn, ref, "") do
    Mint.HTTP.stream_request_body(conn, ref, :eof)
  end

  defp stream_body(conn, ref, body) do
    {chunk, rest} = Mint.HTTP.next_body_chunk(conn, ref, body)

    with {:ok, conn} <- Mint.HTTP.stream_request_body(conn, ref, chunk) do
      stream_body(conn, ref, rest)
    end
  end
end
```

Wins:

- allows to keep `stream_request_body/3` protocol-agnostic
- provides convenient API for the most obvious and straightforward use-case
- keeps the API a low-level and allow caller to have all the control

### DIFF
--- a/.dialyzer_ignore
+++ b/.dialyzer_ignore
@@ -1,5 +1,5 @@
 lib/mint/tunnel_proxy.ex:49
-lib/mint/http1.ex:915
+lib/mint/http1.ex:940
 lib/mint/unsafe_proxy.ex:173
 lib/mint/unsafe_proxy.ex:198
 test/support
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### New features
+
+  * Add `Mint.HTTP.next_body_chunk/3` and `Mint.HTTP.next_body_chunk_size/2` helpers for streaming a request body of arbitrary size while respecting the connection's current send window. In HTTP/2 the chunk size is bounded by the connection-level and per-request flow-control windows; in HTTP/1 it is bounded by the OS socket send buffer (`:sndbuf`) cached at connect time.
+
 ## v1.7.1
 
 ### Bug Fixes and Improvements
--- a/lib/mint/core/conn.ex
+++ b/lib/mint/core/conn.ex
@@ -62,4 +62,6 @@ defmodule Mint.Core.Conn do
   @callback put_proxy_headers(conn(), Mint.Types.headers()) :: conn()
 
   @callback put_log(conn(), boolean()) :: conn()
+
+  @callback next_body_chunk_size(conn(), Types.request_ref()) :: non_neg_integer()
 end
--- a/lib/mint/http.ex
+++ b/lib/mint/http.ex
@@ -623,6 +623,9 @@ defmodule Mint.HTTP do
 
   This function always returns an updated connection to be stored over the old connection.
 
+  When streaming a body of arbitrary size, use `next_body_chunk/3` to split it
+  into chunks that respect the current send window of the connection.
+
   For information about transfer encoding and content length in HTTP/1, see
   `Mint.HTTP1.stream_request_body/3`.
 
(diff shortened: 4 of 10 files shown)


ericmj:
You are treating H2 send window and sndbuf as two implementations of "how big should the next chunk be," but they're values from different layers that serve different purposes.

The H2 send window is a contract with the peer, if we exceed it we get a protocol error, Mint checks it because violating it is a wire-level bug, not for performance or to avoid blocking.

sndbuf is the size of the kernel send buffer. It's not a contract and no error if you "exceed" it. gen_tcp.send just blocks until the kernel drains enough bytes to accept more.

For H1, sndbuf is the buffer's total capacity, not its current free space (unlike H2 windows). Send one sndbuf-sized chunk and it sits in the kernel buffer until the receiver ACKs, if we send a second one immediately and it blocks waiting for ACKs. Chunking by sndbuf only sizes the first send to fit an empty buffer, every send after that depends on how fast ACKs come back.

For H2, the window can easily exceed sndbuf. When the window is bigger, sending up to it fills the kernel buffer and blocks anyway. The H2 code only guarantees protocol compliance, not non-blocking sends.

For these reasons the proposed implementation is a leaky abstraction over H1 and H2. With the change in https://github.com/elixir-mint/mint/pull/483 users can implement streaming and it's up to the user to pick a chunk size that's appropriate for their application's needs.

ericmj:
To elaborate on this a bit more and suggest a hopefully acceptable solution.

The library should not pick the application chunk size. The user should pick a chunk size based on latency, memory, producer shape, backpressure model, and tolerance for blocking. Mint should only expose the protocol constraint. 

Given that, we can add a function `Mint.HTTP.request_body_window(conn, ref) :: :infinity | non_neg_integer()`. This would be the upper limit of what a user is allowed to send, then a user should do:

```elixir
user_chunk_size = 16_384

limit =
  case Mint.HTTP.request_body_window(conn, ref) do
    :infinity -> user_chunk_size
    window -> min(user_chunk_size, window)
  end
```

tank-bohr:
@ericmj  Thank you! I agree the current implementation neither guarantees non-blocking behaviour, nor reflects the protocol requirements. But trying to be both. Pretty confusing. One more attempt #485

PR #485, opened 2026-04-26

tank-bohr:
See https://github.com/elixir-mint/mint/pull/484#issuecomment-4321958172 for details

### DIFF
--- a/lib/mint/http.ex
+++ b/lib/mint/http.ex
@@ -1082,10 +1082,14 @@ defmodule Mint.HTTP do
       for the underlying primitives.
 
     * In HTTP/1, returns `:infinity`. HTTP/1 has no application-level
-      flow-control mechanism: any amount of body data is protocol-valid. The
-      operating-system socket send buffer may still cause `stream_request_body/3`
-      to block once it fills up, but that is a transport concern and not
-      reflected here.
+      flow-control mechanism: any amount of body data is protocol-valid.
+
+  The value returned reflects only the protocol-level flow-control
+  constraint. It does not account for the operating-system socket send
+  buffer: under either protocol, `stream_request_body/3` can still block
+  when that buffer fills up. To bound this behavior, configure
+  `send_timeout` on the socket via `:transport_opts` when establishing the
+  connection (see `Mint.HTTP.connect/4`).
 
   Raises `ArgumentError` if `request_ref` is not associated with an active
   streaming request.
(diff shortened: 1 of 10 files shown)


lib/mint/http.ex
```diff
@@ -1065,6 +1069,53 @@ defmodule Mint.HTTP do
...
+      number of body bytes that can be sent right now without violating flow
+      control. Exceeding this value in a single `DATA` frame would close the
+      connection with a `FLOW_CONTROL_ERROR`. See `Mint.HTTP2.get_window_size/2`
+      for the underlying primitives.
+
+    * In HTTP/1, returns `:infinity`. HTTP/1 has no application-level
+      flow-control mechanism: any amount of body data is protocol-valid. The
+      operating-system socket send buffer may still cause `stream_request_body/3`
+      to block once it fills up, but that is a transport concern and not
+      reflected here.
```
ericmj: This is a good note to make but it's not H1 specific. In general we cannot guarantee non-blocking unless you set `send_timeout` on the socket through `transport_opts`.

lib/mint/http.ex
```diff
@@ -1065,6 +1069,53 @@ defmodule Mint.HTTP do
...
+      end
+
+      defp stream_body(conn, ref, body) do
+        chunk_size = min(Mint.HTTP.request_body_window(conn, ref), byte_size(body))
+        <<chunk::binary-size(chunk_size), rest::binary>> = body
+
+        with {:ok, conn} <- Mint.HTTP.stream_request_body(conn, ref, chunk) do
+          stream_body(conn, ref, rest)
+        end
+      end
```
ericmj, line 1124: There are two issues with this code example.
* We can return 0 from `request_body_window/2` so you can attempt to send an empty binary.
* There is no backpressure, if there is no window available we will be busy looping, consuming CPU resources.

The latter can be hard to express in a code example because there are many different solutions to it depending on context, so maybe we should just call it out?

ericmj (commented):
This looks great, thanks for pushing this through and for sticking with this through multiple revisions.

I left two comments about documentation only.

lib/mint/http.ex
tank-bohr: Fixed

lib/mint/http.ex
tank-bohr: Great catch! Fixed

ericmj (approved):

PR #486, opened 2026-05-04
Don't close the connection on a receive timeout

keathley:
Its often useful to be able to test that a connection has no data waiting on it like `recv(socket, 0, 0)`. This PR handles the timeout error and returns it to the caller, rather than defaulting to closing the connection.


ericmj:
CI found a failing test for this.

Is it clear from the API when Mint closes a connection on errors and when it doesn't or should we document this?

whatyouhide:
To Eric's comment I think yes we should. I feel it would be pretty confusing to not know that, and while I think not closing on timeouts is reasonable I also want to make sure we loudly document that.

keathley:
@whatyouhide That makes sense to me. I should have time to fix tests and update docs later today.

PR #491, opened 2026-07-18
Send origin Host header for plain-HTTP proxied requests

### LINKED ISSUE #446: HTTP_PROXY doesn't set host for a connection, using proxy as host
Reading:
https://hexdocs.pm/mint/Mint.HTTP.html#connect/4-proxying

A minimal example, using [mitmproxy](https://mitmproxy.org/):

```elixir
proxy_details = {:http, "127.0.0.1", 9055, []}
{:ok, conn} = Mint.HTTP.connect(:http, "example.com", 80, proxy: proxy_details)
{:ok, conn, request_ref} = Mint.HTTP.request(conn, "GET", "/", [], nil)
```

The raw request seems to be:

```
GET / HTTP/1.1
host: 127.0.0.1:9055
user-agent: mint/1.5.1
```

When using hackney, this seems to work:

```elixir
# Set up the proxy options
proxy_opts = [
  proxy: {"127.0.0.1", 9055}
]

# Make the request using hackney
:hackney.request(:get, "http://example.com", [], "", proxy_opts)
```

Request:
```
GET / HTTP/1.1
Host: example.com
User-Agent: hackney/1.20.1
```

(HTTPoison also works, `HTTPoison.get("http://example.com", [], [{:proxy, string_url()}])`)

Leonard013:
## Fix `Host` header for plain-HTTP proxied requests

Fixes #446.

### Problem

When connecting to an origin server through a plain-HTTP proxy —
`Mint.HTTP.connect(:http, "example.com", 80, proxy: {:http, "127.0.0.1", 9055, []})` —
Mint sends the **proxy's** address as the `Host` header instead of the origin's:

```
GET http://example.com/ HTTP/1.1
host: 127.0.0.1:9055        <-- wrong: this is the proxy, not the origin
user-agent: mint/1.9.3
```

hackney and HTTPoison send `Host: example.com` for the identical proxied
request. Per RFC 7230 §5.4 the `Host` header must identify the origin server,
so Mint's behavior breaks virtual-hosted origins, proxy logging, and any server
behind the proxy that routes on `Host`.

### Root cause

`Mint.UnsafeProxy` (the plain-HTTP, non-tunnel proxy) holds an `Mint.HTTP1`
connection whose socket points at the *proxy*. In `Mint.UnsafeProxy.request/5`
the request path is correctly rewritten to the absolute origin URI, but no
`Host` header is set, so it falls through to `Mint.HTTP1`, whose
`add_default_headers/2` fills in `Host` from *its* connection — i.e. the proxy's
host and port (`lib/mint/http1.ex:1170`, `default_host_header/1` at
`lib/mint/http1.ex:1174`). `UnsafeProxy` is the only layer that knows the
origin, so the origin `Host` has to be set there.

### Fix

`lib/mint/unsafe_proxy.ex` — `request/5` now sets the `Host` header to the
origin server (host, plus port when it is not the scheme default) before
delegating to `Mint.HTTP1`. Because `Mint.HTTP1` uses `put_new` for its default
`Host`, the origin value is left untouched. A caller-supplied `Host` header
(matched case-insensitively) is still respected, matching the direct-connection
behavior.

### Tests

New `test/mint/unsafe_proxy_host_header_test.exs`. It uses a local one-shot TCP
server as the "proxy" and inspects the raw request bytes, so it needs no live
proxy or internet connection and runs in the default suite (it is a separate
module so it isn't excluded by `unsafe_proxy_test.exs`'s `:proxy` moduletag).
Covers: origin `Host` for a default port, origin `host:port` for a non-default
port, and that a caller-supplied `Host` is not overridden.

### Verification

Before (on `main`):

```
GET http://example.com/ HTTP/1.1
host: localhost:60234
user-agent: mint/1.9.3
```

After:

```
GET http://example.com/ HTTP/1.1
user-agent: mint/1.9.3
host: example.com
```

Commands:

```
$ mix test test/mint/unsafe_proxy_host_header_test.exs
...
3 passed

# whole offline suite unaffected
$ mix test --exclude requires_internet_connection
304 passed, 51 excluded

$ mix format --check-formatted   # clean
$ mix compile --warnings-as-errors   # clean
$ mix dialyzer   # done (passed successfully)
```

The new tests fail before the fix (`host: localhost:<port>` / `host:
example.com:8080`) and pass after.

### Compatibility

No public API change. Only the `Host` header for plain-HTTP proxied requests is
affected, bringing it in line with RFC 7230 §5.4 and with hackney/HTTPoison.
Callers that explicitly pass their own `Host` header are unaffected. Tunnel
(CONNECT) proxying and direct connections are untouched.

---

*AI-authored and AI-reviewed; not yet human-reviewed.*

### DIFF
--- a/lib/mint/unsafe_proxy.ex
+++ b/lib/mint/unsafe_proxy.ex
@@ -81,7 +81,7 @@ defmodule Mint.UnsafeProxy do
         body \\ nil
       ) do
     path = request_line(conn, path)
-    headers = headers ++ conn.proxy_headers
+    headers = put_new_host_header(headers, conn) ++ conn.proxy_headers
 
     case module.request(state, method, path, headers, body) do
       {:ok, state, request} -> {:ok, %{conn | state: state}, request}
@@ -203,5 +203,29 @@ defmodule Mint.UnsafeProxy do
   @impl true
   def request_body_window(%__MODULE__{module: module, state: state}, ref) do
     module.request_body_window(state, ref)
+  end
+
+  # When proxying over plain HTTP, the request is sent to the proxy but its Host
+  # header must identify the origin server, not the proxy (RFC 7230, sec. 5.4).
+  # Mint.HTTP1 would otherwise default the Host header to the proxy's address,
+  # since its connection points at the proxy. We set it here (unless the caller
+  # already provided one) so Mint.HTTP1's `put_new` leaves the origin's Host in
+  # place.
+  defp put_new_host_header(headers, conn) do
+    if Enum.any?(headers, fn {name, _value} -> String.downcase(name, :ascii) == "host" end) do
+      headers
+    else
+      [{"Host", host_header_value(conn)} | headers]
+    end
+  end
+
+  # Mirrors the default-port handling in Mint.HTTP1: omit the port when it is the
+  # default for the scheme.
+  defp host_header_value(%UnsafeProxy{scheme: scheme, hostname: hostname, port: port}) do
+    if URI.default_port(Atom.to_string(scheme)) == port do
+      hostname
+    else
+      "#{hostname}:#{port}"
+    end
   end
 end
--- a/lib/mint/unsafe_proxy.ex
+++ b/lib/mint/unsafe_proxy.ex
@@ -212,7 +212,7 @@ defmodule Mint.UnsafeProxy do
   # already provided one) so Mint.HTTP1's `put_new` leaves the origin's Host in
   # place.
   defp put_new_host_header(headers, conn) do
-    if Enum.any?(headers, fn {name, _value} -> String.downcase(name, :ascii) == "host" end) do
+    if Enum.any?(headers, fn {name, _value} -> Mint.Core.Headers.lower_raw(name) == "host" end) do
       headers
     else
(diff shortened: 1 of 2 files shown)


lib/mint/unsafe_proxy.ex
```diff
@@ -204,4 +204,28 @@ defmodule Mint.UnsafeProxy do
...
   end
+
+  # When proxying over plain HTTP, the request is sent to the proxy but its Host
+  # header must identify the origin server, not the proxy (RFC 7230, sec. 5.4).
+  # Mint.HTTP1 would otherwise default the Host header to the proxy's address,
+  # since its connection points at the proxy. We set it here (unless the caller
+  # already provided one) so Mint.HTTP1's `put_new` leaves the origin's Host in
+  # place.
+  defp put_new_host_header(headers, conn) do
+    if Enum.any?(headers, fn {name, _value} -> String.downcase(name, :ascii) == "host" end) do
```
ericmj:
```suggestion
    if Enum.any?(headers, fn {name, _value} -> Mint.Core.Headers.lower_raw(name) == "host" end) do
```

ericmj:
Thank you!

PR #495, opened 2026-07-21
Return a tunnel timeout error when the CONNECT deadline elapses

ericmj:
A proxy that dribbles the CONNECT response until the tunnel deadline passes mid-loop makes `receive_response/3` compute a negative timeout for its `after` clause, which raises `ErlangError :timeout_value` in the calling process instead of returning `{:error, %Mint.HTTPError{reason: {:proxy, :tunnel_timeout}}}`. This clamps the timeout to 0 so an elapsed deadline yields the documented error, with offline regression tests (the crash reproduces deterministically with an already-elapsed deadline).

Also documents `:tunnel_timeout` in the Proxying section, which was previously only mentioned in the `Mint.HTTPError` docs.

### DIFF
--- a/lib/mint/http.ex
+++ b/lib/mint/http.ex
@@ -287,6 +287,15 @@ defmodule Mint.HTTP do
   uses TLS and the TLS session to the host is nested inside it. *HTTPS proxies
   for HTTPS connections are available since v1.10.0*.
 
+  The `opts` in the `:proxy` tuple are the options of the connection to the proxy
+  itself and support the same options as `connect/4`. Tunnel proxies additionally
+  support:
+
+    * `:tunnel_timeout` - the maximum time (in milliseconds) to wait for the proxy
+      to reply to the `CONNECT` request. If the tunnel is not established within
+      this time, `connect/4` returns a `Mint.HTTPError` with reason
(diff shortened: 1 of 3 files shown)


lib/mint/http.ex
```diff
@@ -287,6 +287,15 @@ defmodule Mint.HTTP do
...
   for HTTPS connections are available since v1.10.0*.
 
+  The `opts` in the `:proxy` tuple are the options of the connection to the proxy
+  itself and support the same options as `connect/4`. Tunnel proxies additionally
+  support:
+
+    * `:tunnel_timeout` - the maximum time (in milliseconds) to wait for the proxy
+      to reply to the `CONNECT` request. If the tunnel is not established within
+      this time, `connect/4` returns a `Mint.HTTPError` with reason
+      `{:proxy, :tunnel_timeout}`. Defaults to `30_000` (30 seconds).
```
whatyouhide, line 297: This is missing the `*Available since vX.X.X*` notice.

whatyouhide (approved):

lib/mint/http.ex
ericmj: This option has been available since 0.1.0, it was just not documented.

Issue #452, opened 2024-10-18
Permit processing HTTP1 headers as they arrive


nsweeting:
I'm currently trying to implement a HTTP-based database driver using mint. One of the features that the DB supports is "progress headers". The headers are continually sent to the client to provide feedback on the query progress. Unfortunately I don't seem to be able to access these until the query is fully executed as all the headers come at once via `{:headers, request_ref, headers}` - and as noted in the docs "A single `:headers` response will come after the `:status` response.". I can confirm that the headers are being sent as the query progresses went the same request is sent with curl.

Is there a reason why mint couldn't provide access to these headers as they arrive instead of accumulating them?

whatyouhide:
Hi @nsweeting are you using HTTP/1.1 or HTTP/2?

Malien:
This would either be a breaking change, or would need to opted into by a flag/config, as the docs make a guarantee that:
> A single :headers response will come after the :status response.

PR #499, opened 2026-08-10
Support processing HTTP/1.1 headers as they arrive

gdwoolbert3:
## Introduction

This is my attempt at implementing the feature requested in [Issue 452](https://github.com/elixir-mint/mint/issues/452).

The motivation for the change was the `X-ClickHouse-Progress` header system used by the [ClickHouse HTTP interface](https://clickhouse.com/docs/concepts/features/interfaces/http) (which I strongly suspect was the reason behind the aforementioned issue).

## Implementation

This PR adds a new `:stream_headers` option for HTTP/1.1 connections that allows headers and trailers to be emitted as responses as soon as they're available instead of buffering them until they've all arrived. A few nuances about enabling this feature:
- All _complete_ headers from a single chunk are emitted at the same time
- It's theoretically possible to receive **NO** `:headers` responses for a given request whereas, when not streaming headers, an empty list would be emitted. This change is explicitly laid out in the `Mint.HTTP` documentation.

This feature is opt-in, meaning there should be no changes for existing users.

## Tesing

All existing tests pass and new stream headers tests were added to the HTTP/1.1 connection test module.

Open to any and all feedback 😃

### DIFF
--- a/lib/mint/http1.ex
+++ b/lib/mint/http1.ex
@@ -104,6 +104,7 @@ defmodule Mint.HTTP1 do
     :case_sensitive_headers,
     :skip_target_validation,
     :max_header_list_size,
+    :stream_headers,
     requests: :queue.new(),
     state: :closed,
     buffer: "",
@@ -149,6 +150,11 @@ defmodule Mint.HTTP1 do
     * `:max_header_list_size` - (`t:pos_integer/0` or `:infinity`) the maximum number of
       bytes allowed in a response header section or chunked trailer section. This includes
       header names, values, and line delimiters. Defaults to 256 KiB. *Available since 1.9.2*.
+
+    * `:stream_headers` - (`t:boolean/0`) if set to `true`, response headers and trailer headers
+      will be emitted as they are parsed, rather than buffered until the complete header section
+      is received. When enabled, you may receive multiple `{:headers, ref, headers}` responses
+      for a single request. Defaults to `false`. *Available since v1.10.0*.
 
   """
   @spec connect(Types.scheme(), Types.address(), :inet.port_number(), keyword()) ::
@@ -238,6 +244,7 @@ defmodule Mint.HTTP1 do
         case_sensitive_headers: Keyword.get(opts, :case_sensitive_headers, false),
         skip_target_validation: Keyword.get(opts, :skip_target_validation, false),
         max_header_list_size: max_header_list_size,
+        stream_headers: Keyword.get(opts, :stream_headers, false),
         optional_responses: validate_optional_response_values(opts)
       }
 
@@ -758,7 +765,12 @@ defmodule Mint.HTTP1 do
       {:ok, :eof, rest} ->
         case add_header_bytes(conn, request, byte_size(data) - byte_size(rest)) do
           {:ok, request} ->
-            responses = [{:headers, request.ref, Enum.reverse(headers)} | responses]
+            responses =
+              case {conn.stream_headers, headers} do
+                {true, []} -> responses
+                {_, headers} -> [{:headers, request.ref, Enum.reverse(headers)} | responses]
+              end
+
             request = %{request | state: :body, headers_buffer: [], headers_size: 0}
             conn = %{conn | buffer: "", request: request}
             decode(:body, conn, rest, responses)
@@ -770,7 +782,19 @@ defmodule Mint.HTTP1 do
       :more ->
         case check_header_section_size(conn, request.headers_size + byte_size(data)) do
           :ok ->
-            request = %{request | headers_buffer: headers}
+            {responses, headers_buffer} =
+              cond do
+                not conn.stream_headers ->
+                  {responses, headers}
+
+                headers != [] ->
+                  {[{:headers, request.ref, Enum.reverse(headers)} | responses], []}
+
+                true ->
+                  {responses, []}
+              end
+
+            request = %{request | headers_buffer: headers_buffer}
             conn = %{conn | buffer: data, request: request}
             {:ok, conn, responses}
 
@@ -950,7 +974,24 @@ defmodule Mint.HTTP1 do
       :more ->
         case check_header_section_size(conn, conn.request.headers_size + byte_size(data)) do
           :ok ->
-            request = %{conn.request | body: {:chunked, :trailer}, headers_buffer: headers}
+            {responses, headers_buffer} =
+              cond do
+                not conn.stream_headers ->
+                  {responses, headers}
+
+                headers != [] ->
+                  responses =
+                    headers
+                    |> Headers.remove_unallowed_trailer()
+                    |> add_trailer_headers(conn.request.ref, responses)
+
+                  {responses, []}
+
+                true ->
+                  {responses, []}
+              end
+
+            request = %{conn.request | body: {:chunked, :trailer}, headers_buffer: headers_buffer}
             conn = %{conn | buffer: data, request: request}
             {:ok, conn, responses}
 
--- a/lib/mint/http1.ex
+++ b/lib/mint/http1.ex
@@ -751,7 +751,7 @@ defmodule Mint.HTTP1 do
   end
 
   defp decode_headers(conn, request, data, responses, headers) do
-    case Response.decode_header(data) do
+    case do_decode_header(data, conn.stream_headers) do
       {:ok, {name, value}, rest} ->
         headers = [{name, value} | headers]
 
@@ -942,7 +942,7 @@ defmodule Mint.HTTP1 do
   end
 
   defp decode_trailer_headers(conn, data, responses, headers) do
-    case Response.decode_header(data) do
+    case do_decode_header(data, conn.stream_headers) do
       {:ok, {name, value}, rest} ->
         case add_header_bytes(conn, conn.request, byte_size(data) - byte_size(rest)) do
           {:ok, request} ->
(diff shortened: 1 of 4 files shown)


ericmj:
The option and response shape make sense, but there’s one parsing case to fix before merging. If a socket message ends exactly after a complete header line, such as `X-ClickHouse-Progress: ...\r\n`, `HTTP1.stream/2` buffers the header and returns no response until another byte arrives. This leaves progress updates one socket message behind, and a lone update won’t be emitted until the next header or the end of the section.

Could you add tests where a regular header and a trailer end at that boundary, with no byte from the following line, and update the parser to emit them immediately?

gdwoolbert3:
@ericmj nice catch! If I'm understanding correctly, this appears to be a quirk of the underlying `:erlang.decode_packet/3` function:

```
iex> :erlang.decode_packet(:httph_bin, "X-ClickHouse-Progress: foo\r\n", [])
{:more, :undefined}

iex> :erlang.decode_packet(:httph_bin, "X-ClickHouse-Progress: foo\r\nX", [])
{:ok, {:http_header, 0, "X-Clickhouse-Progress", "X-ClickHouse-Progress", "foo"}, "X"}
```

With that being said, the ideal fix isn't particularly clear to me. If I'm reading RFC 7230 right, I _think_ we could naively check that the incoming data ends with `\r\n` and then do something to force the header to parse (ie appending a dummy byte), but that doesn't feel like the cleanest solution.

Do you have any suggestions? If not, I'll go with the aforementioned approach.

On a related note, do you have insight into whether or not the above behavior is intentional?

wojtekmach:
one reason might be HTTP1 line folding, a header could have spanned multiple lines, https://www.rfc-editor.org/info/rfc9112/#section-5.2,

```elixir
iex> :erlang.decode_packet(:httph_bin, "header: chunk1\r\n chunk2\r\n\r\n", [])
{:ok, {:http_header, 0, "Header", "header", "chunk1\r\n chunk2"}, "\r\n"}
```

and so when getting `"header: chunk1\r\n"`, returning :more makes sense. That particular behaviour is obsolete in the spec however. And so yeah, it would be nice if we've gotten `{:ok, {:http_header, ...}, ""}` instead, but that'd probably require adding an option to packet parser and that probably has very high barrier to entry.

ericmj:
Let's retry the `:erlang.decode_packet` call with `data <> sentinel` when it returns `:more`, `stream_headers` is enabled, and `data` ends in LF. Where `sentinel` is a single byte that doesn't trigger the legacy line folding.

gdwoolbert3:
[Pushed](ff47c8315322b9b2556e8a02e8aab3c9589d1b22) the fix recommended by @ericmj.

As it turns out, `:erlang.decode_packet/3` allows LF in addition to CRLF as a header sep:

```
iex> :erlang.decode_packet(:httph_bin, "X-ClickHouse-Progress: foo\r\n", [])
{:more, :undefined}
iex> :erlang.decode_packet(:httph_bin, <<"X-ClickHouse-Progress: foo\r\n", 0>>, [])
{:ok, {:http_header, 0, "X-Clickhouse-Progress", "X-ClickHouse-Progress", "foo"}, <<0>>}
iex> :erlang.decode_packet(:httph_bin, <<"X-ClickHouse-Progress: foo\n", 0>>, [])
{:ok, {:http_header, 0, "X-Clickhouse-Progress", "X-ClickHouse-Progress", "foo"}, <<0>>}
```

I kept this behavior in order to maintain parity with the underlying functionality, but I'm happy to update at your discretion.

lib/mint/http1.ex
```diff
@@ -1004,6 +1004,25 @@ defmodule Mint.HTTP1 do
...
+
+  defp do_decode_header(data, true) do
+    # By default, :erlang.decode_packet/3 asks for more data when a packet
+    # containing a full header ends with a line feed (likely to handle line
+    # folding). If we get a :more response on a packet that ends with a line
+    # feed, we append a sentinel byte and attempt to decode again.
+    with :more <- Response.decode_header(data) do
+      data_size = byte_size(data)
+
+      with <<_::binary-size(data_size - 1), 10>> <- data,
```
ericmj:
```suggestion
      with <<_::binary-size(data_size - 1), ?\n>> <- data,
```

lib/mint/http1.ex
```diff
@@ -963,6 +1004,25 @@ defmodule Mint.HTTP1 do
...
+           {:ok, {name, value}, rest} <- Response.decode_header(<<data::binary, 0>>) do
+        {:ok, {name, value}, binary_part(rest, 0, byte_size(rest) - 1)}
+      else
+        _ -> :more
```
ericmj: `MIX_ENV=test mix compile --force --warnings-as-errors` fails because `data_size` is not pinned in the bitstring size expression. This version also preserves the retry error and removes the added byte only when the remainder matches the sentinel.

```suggestion
      with <<_::binary-size(^data_size - 1), ?\n>> <- data do
        case Response.decode_header(<<data::binary, 0>>) do
          {:ok, {name, value}, <<0>>} -> {:ok, {name, value}, ""}
          result -> result
        end
      else
        _ -> :more
```

lib/mint/http1.ex
```diff
@@ -963,6 +1004,25 @@ defmodule Mint.HTTP1 do
     end
   end
 
+  defp do_decode_header(data, false), do: Response.decode_header(data)
+
+  defp do_decode_header(data, true) do
```
ericmj:
```suggestion
  defp do_decode_header(data, true = _stream_headers) do
```

test/mint/http1/conn_test.exs
```diff
@@ -1273,6 +1273,167 @@ defmodule Mint.HTTP1Test do
...
+      {:ok, conn, ref} = HTTP1.request(conn, "GET", "/", [], nil)
+
+      assert {:ok, conn, [_status]} =
+               HTTP1.stream(conn, {:tcp, conn.socket, "HTTP/1.1 200 OK\r\n"})
+
+      # Send a complete header ending exactly at packet boundary (no subsequent bytes)
+      assert {:ok, conn, responses} =
+               HTTP1.stream(conn, {:tcp, conn.socket, "X-Progress: 50\r\n"})
+
+      assert [{:headers, ^ref, [{"x-progress", "50"}]}] = responses
```
ericmj: Use two headers here while keeping the final LF at the packet boundary. This covers the recursive path where the first header has real lookahead and only the final header uses the sentinel.

```suggestion
      assert {:ok, conn, responses} =
               HTTP1.stream(
                 conn,
                 {:tcp, conn.socket, "X-Progress: 50\r\nX-Other: value\r\n"}
               )

      assert [{:headers, ^ref, [{"x-progress", "50"}, {"x-other", "value"}]}] = responses
```

lib/mint/http1.ex
```diff
@@ -1004,6 +1004,25 @@ defmodule Mint.HTTP1 do
     end
   end
 
+  defp do_decode_header(data, false), do: Response.decode_header(data)
```
ericmj:
```suggestion
  defp do_decode_header(data, false = _stream_headers), do: Response.decode_header(data)
```

ericmj (changes requested):

lib/mint/http1.ex
gdwoolbert3: 5940b9a68d479de238e34fb4f3dc4e1d55ed4140

lib/mint/http1.ex
gdwoolbert3: Ahh, interesting! I didn't know about the pin being preferred there.

My initial assumption was that, if the initial decode didn't fail, it wouldn't fail with the sentinel bit. That said, your approach is much more explicit.
5940b9a68d479de238e34fb4f3dc4e1d55ed4140

test/mint/http1/conn_test.exs
gdwoolbert3: 5940b9a68d479de238e34fb4f3dc4e1d55ed4140

lib/mint/http1.ex
gdwoolbert3: 5940b9a68d479de238e34fb4f3dc4e1d55ed4140

lib/mint/http1.ex
gdwoolbert3: I made one small tweak here in regards to the usage of `with` and `case` but the control flow is the same:
9c223c686ddb61802f9e2f16fbd00b7f74353d2d

ericmj (approved):

lib/mint/http1.ex
```diff
@@ -744,7 +751,7 @@ defmodule Mint.HTTP1 do
   end
 
   defp decode_headers(conn, request, data, responses, headers) do
-    case Response.decode_header(data) do
+    case do_decode_header(data, conn.stream_headers) do
```
whatyouhide: **Nit**: the `do_` prefix has disappeared from Elixir conventions years ago. Can we rename this to `decode_header/2` or `decode_single_header/2` if that's taken or something like that?

lib/mint/http1.ex
```diff
@@ -758,7 +765,12 @@ defmodule Mint.HTTP1 do
       {:ok, :eof, rest} ->
         case add_header_bytes(conn, request, byte_size(data) - byte_size(rest)) do
           {:ok, request} ->
-            responses = [{:headers, request.ref, Enum.reverse(headers)} | responses]
+            responses =
+              case {conn.stream_headers, headers} do
+                {true, []} -> responses
+                {_, headers} -> [{:headers, request.ref, Enum.reverse(headers)} | responses]
+              end
```
whatyouhide: I think this reads significantly better as:


```suggestion
              if conn.stream_headers and headers == [] do
                responses
              else
                [{:headers, request.ref, Enum.reverse(headers)} | responses]
              end
```

lib/mint/http1.ex
gdwoolbert3: c2036c34cf10912bd1776d0f1f0868ab2000fc81

lib/mint/http1.ex
gdwoolbert3: c2036c34cf10912bd1776d0f1f0868ab2000fc81

gdwoolbert3:
@whatyouhide apologies, I had to update the line reference in the dialyzer ignore file. I suspect that the workflow should pass now

whatyouhide:
Thank you @gdwoolbert3! 🫶

PR #500, opened 2026-08-10
Sort generated functions for deterministic builds

fhunleth:
When rebuilding frame.ex without any source code changes, the resulting
.beam file changed. This was due to functions being reordered in the
compiled .beam file depending on the ordering of several maps. The fix
is to sort the maps to make the order deterministic.

### DIFF
--- a/lib/mint/http2/frame.ex
+++ b/lib/mint/http2/frame.ex
@@ -37,7 +37,7 @@ defmodule Mint.HTTP2.Frame do
 
   @spec inspect(tuple()) :: String.t()
 
-  for {type, _code} <- @types do
+  for {type, _code} <- Enum.sort(@types) do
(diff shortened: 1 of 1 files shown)


fhunleth:
Just to provide some more context here, it's pretty easy to see that something changed in the build products with Nerves since firmware images get referred to by their checksums. This was found by building the same project twice and noticing an unexpected change to the final firmware. On inspection, this was a harmless change, but it's nice when you don't have to look through a .beam file to figure that out. Plus getting rid of the change reduces the size of delta firmware updates by a little bit. I'm hoping that deterministic builds are helpful to others too.

fhunleth:
Last note, when I ran the unit tests on macOS, I had 16 failures, but those tests failed before and after the change so I think I'm not running the tests correctly.

whatyouhide:
This is great, thank you Frank 🫶

PR #502, opened 2026-09-04
Close TCP sockets on errors in HTTP/1

### LINKED ISSUE #501: TCP sockets get leaked on certain failures
Hi, thank you for creating and maintaining this library! ❤️ 

Claude has a found a bug in the way that some failed responses are handled. In three tested cases (invalid status line, duplicate content-length header, sending less content than content-length header says), Mint correctly reports the error but it probably forgets to close the TCP socket. It then says that the `state` is `:closed` and therefore calling `Mint.HTTP.close/1` is a no-op and does not close the socket. These open sockets can then accumulate until resources are exhausted.

Note that I discovered this bug via Req which transitively depends on Mint.

<details>
<summary>An AI-generated reproducer (foldable section).</summary>

```elixir
# A response that fails to parse (or is cut short by the peer) leaks its TCP
# socket: Mint marks the conn `state: :closed` without closing the socket, and
# `Mint.HTTP.close/1` is a no-op for that state, so the socket can never be
# released — even though the client below always calls close/1.
#
# Run: elixir mint_leak_repro.exs
Mix.install([{:mint, "~> 1.9.3"}])

defmodule LeakRepro do
  def start_server(response, close_after_reply?) do
    {:ok, lsock} = :gen_tcp.listen(0, [:binary, active: false])
    {:ok, port} = :inet.port(lsock)
    spawn_link(fn -> accept_loop(lsock, response, close_after_reply?) end)
    port
  end

  defp accept_loop(lsock, response, close_after_reply?) do
    {:ok, sock} = :gen_tcp.accept(lsock)

    spawn_link(fn ->
      {:ok, _request} = :gen_tcp.recv(sock, 0, 5000)
      :ok = :gen_tcp.send(sock, response)
      if close_after_reply?, do: :gen_tcp.close(sock), else: Process.sleep(:infinity)
    end)

    accept_loop(lsock, response, close_after_reply?)
  end

  # One GET on a fresh connection; always closes the conn before returning.
  def get(port) do
    {:ok, conn} = Mint.HTTP.connect(:http, "127.0.0.1", port, mode: :passive)
    {:ok, conn, _ref} = Mint.HTTP.request(conn, "GET", "/", [], nil)
    {result, conn} = recv_response(conn, nil)
    {:ok, _conn} = Mint.HTTP.close(conn)
    result
  end

  # Reads the whole response; returns the status, or the error reason.
  defp recv_response(conn, status) do
    case Mint.HTTP.recv(conn, 0, 2000) do
      {:ok, conn, responses} ->
        status =
          Enum.find_value(responses, status, fn
            {:status, _ref, code} -> code
            _other -> nil
          end)

        if Enum.any?(responses, &match?({:done, _ref}, &1)),
          do: {status, conn},
          else: recv_response(conn, status)

      {:error, conn, error, _responses} ->
        # assert that in case of error, the state is always :closed
        :closed = conn.state

        {error.reason, conn}
    end
  end

  def sockets_owned_by_me do
    Enum.count(:erlang.ports(), fn p ->
      :erlang.port_info(p, :name) == {:name, ~c"tcp_inet"} and
        :erlang.port_info(p, :connected) == {:connected, self()}
    end)
  end
end

cases = [
  {"valid response (control)", "HTTP/1.1 200 OK\r\ncontent-length: 2\r\n\r\nok", false},
  {"garbage status line", "HELLO WORLD\r\n\r\n", false},
  {"duplicate content-length",
   "HTTP/1.1 200 OK\r\ncontent-length: 2\r\ncontent-length: 3\r\n\r\nok", false},
  {"body shorter than content-length, peer closes",
   "HTTP/1.1 200 OK\r\ncontent-length: 100\r\n\r\nshort", true}
]

IO.puts("test cases (10 requests per case):")

for {label, response, close_after_reply?} <- cases do
  port = LeakRepro.start_server(response, close_after_reply?)
  baseline = LeakRepro.sockets_owned_by_me()
  results = for _ <- 1..10, do: LeakRepro.get(port)
  leaked = LeakRepro.sockets_owned_by_me() - baseline

  IO.puts(
    "  #{String.pad_trailing(label, 46)} -> #{String.pad_trailing(inspect(Enum.uniq(results)), 40)} -> sockets leaked: #{leaked}"
  )
end
```
</details>

On my machine (and I believe it should anywhere), the reproducer outputs:
```
test cases (10 requests per case):
  valid response (control)                       -> [200]                                    -> sockets leaked: 0
  garbage status line                            -> [:invalid_status_line]                   -> sockets leaked: 10
  duplicate content-length                       -> [:more_than_one_content_length_header]   -> sockets leaked: 10
  body shorter than content-length, peer closes  -> [:closed]                                -> sockets leaked: 10
```

Tested on Mint 1.9.3, OTP 29, Elixir 1.20

xxdavid:
In some cases, when handling a failure, Mint would only set the conn state to `:closed` but would not close the underlying socket. These sockets stay open until the calling process that owns them exits, and may therefore accumulate until resources are exhausted.

Every relevant `%{conn | state: :closed}` inside `Mint.HTTP1` is replaced with `internal_close` that in addition to setting the state also closes the socket.

Tests that asserted closed conn now also assert that the socket is closed. New tests that demonstrate the problem are added. They would fail without the fix and they pass with the fix in place.

Fixes #501.

### DIFF
--- a/.dialyzer_ignore
+++ b/.dialyzer_ignore
@@ -1,5 +1,5 @@
 lib/mint/tunnel_proxy.ex:50
-lib/mint/http1.ex:1063
+lib/mint/http1.ex:1066
 lib/mint/unsafe_proxy.ex:173
 lib/mint/unsafe_proxy.ex:198
 test/support
--- a/lib/mint/http1.ex
+++ b/lib/mint/http1.ex
@@ -363,7 +363,8 @@ defmodule Mint.HTTP1 do
       end
     else
       {:error, %TransportError{reason: :closed} = error} ->
-        {:error, %{conn | state: :closed}, error}
+        conn = internal_close(conn)
+        {:error, conn, error}
 
       {:error, %error_module{} = error} when error_module in [HTTPError, TransportError] ->
         {:error, conn, error}
@@ -432,7 +433,8 @@ defmodule Mint.HTTP1 do
         {:ok, conn}
 
       {:error, %TransportError{reason: :closed} = error} ->
-        {:error, %{conn | state: :closed}, error}
+        conn = internal_close(conn)
+        {:error, conn, error}
 
       {:error, error} ->
         {:error, conn, error}
@@ -465,7 +467,8 @@ defmodule Mint.HTTP1 do
         {:ok, conn}
 
       {:error, %TransportError{reason: :closed} = error} ->
-        {:error, %{conn | state: :closed}, error}
+        conn = internal_close(conn)
+        {:error, conn, error}
 
       {:error, error} ->
         {:error, conn, error}
@@ -546,13 +549,13 @@ defmodule Mint.HTTP1 do
         {:ok, conn, Enum.reverse(responses)}
 
       {:error, conn, reason, responses} ->
-        conn = put_in(conn.state, :closed)
+        conn = internal_close(conn)
         {:error, conn, reason, responses}
     end
   end
 
   defp handle_close(%__MODULE__{request: request} = conn) do
-    conn = put_in(conn.state, :closed)
+    conn = internal_close(conn)
     conn = request_done(conn)
 
     if request && request.body == :until_closed do
--- a/test/mint/http1/conn_test.exs
+++ b/test/mint/http1/conn_test.exs
@@ -1,7 +1,7 @@
 defmodule Mint.HTTP1Test do
   use ExUnit.Case, async: true
 
-  alias Mint.{HTTPError, HTTP1, HTTP1.TestServer}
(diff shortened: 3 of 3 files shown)


whatyouhide (approved):
Lovely!

whatyouhide:
@xxdavid can you look at the Dialyzer failure in CI?

xxdavid:
Ah, running Dialyzer did not cross my mind. Fortunately, it was just a stale line number in the ignore list. Thanks for the review!

ericmj:
Thank you! 💜

PR #505, opened 2026-09-20
Check HTTP/2 response bodies against the content-length header

ericmj:
The `content-length` header of an HTTP/2 response was never compared with the DATA frames, so a body that was cut short, or that kept going past the declared length, was delivered as a complete response. Non-numeric and conflicting values were passed through as well.

Each stream now records the request method and the declared length and counts the body bytes it receives. A body that goes past the declared length, or that ends at a different length through END_STREAM on DATA, on the headers or on trailers, is a stream error with PROTOCOL_ERROR. [RFC 9113 8.1.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.1.1) makes it malformed: "A request or response is also malformed if the value of a content-length header field does not equal the sum of the DATA frame payload lengths that form the content, unless the message is defined as having no content", and malformed responses MUST be treated as a stream error of type PROTOCOL_ERROR.

Identical duplicate `content-length` headers are accepted, differing values fail with `:more_than_one_content_length_header`, and values that aren't digits fail with `{:invalid_content_length_header, value}`.

Responses to HEAD, and 204 and 304 responses, are the cases [RFC 9110 6.4.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-6.4.1) says never carry content. They're tracked with a length of zero rather than exempted, so a DATA frame on one is a stream error instead of body data, whatever the `content-length` header says. 2xx responses to CONNECT are exempt from the comparison because they carry tunnel data.

One gap while field names go unvalidated: `content_length/1` matches `{"content-length", value}`, so a server sending `Content-Length` skips the check entirely. Uppercase field names are already prohibited by RFC 9113 8.2.1 and #504 rejects them, so this closes once that lands.


lib/mint/http2.ex
```diff
@@ -378,6 +378,12 @@ defmodule Mint.HTTP2 do
       for example because it contains control characters. `name` is the name of the
       header and `value` is the invalid value.
 
+    * `{:invalid_content_length_header, value}` - when the `content-length` header of a
+      response is not a non-negative integer. `value` is the received value.
+
+    * `:more_than_one_content_length_header` - when a response contains `content-length`
```
whatyouhide: Should we consider calling this `:disagreeing_content_length_headers`?

lib/mint/http2.ex
```diff
@@ -2179,6 +2214,54 @@ defmodule Mint.HTTP2 do
...
+
+  defp content_length(headers) do
+    case for {"content-length", value} <- headers, do: value do
+      [] ->
+        {:ok, nil}
+
+      [value | rest] ->
+        cond do
+          Enum.any?(rest, &(&1 != value)) -> {:error, :more_than_one_content_length_header}
+          not digits?(value) -> {:error, {:invalid_content_length_header, value}}
```
whatyouhide: Shoud we consider a new `Mint.ParsingTools` module (I would love to avoid the name `Utils` 😄), so that we can extract some of these things there? We already have `Mint.HTTP1.Parse.only_digits?/1` for example.

whatyouhide (approved):
Left two small comments but logic looks great.

PR #506, opened 2026-09-20
Ignore HTTP/2 WINDOW_UPDATE frames on closed streams

ericmj:
A WINDOW_UPDATE for a stream Mint no longer tracks made `stream/2` return `{:error, conn, {:stream_not_found, id}, responses}`, so the caller got an error for a frame that's part of normal operation. It arrives whenever the server's flow control update crosses the client's RST_STREAM, or follows the server's own END_STREAM.

[RFC 9113 6.9](https://www.rfc-editor.org/rfc/rfc9113.html#section-6.9): "WINDOW_UPDATE can be sent by a peer that has sent a frame with the END_STREAM flag set. This means that a receiver could receive a WINDOW_UPDATE frame on a stream in a 'half-closed (remote)' or 'closed' state. A receiver MUST NOT treat this as an error (see [Section 5.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-5.1))." The frame is now logged at debug and dropped, which is what DATA, HEADERS and RST_STREAM on a closed stream already did.

PUSH_PROMISE on a closed stream still returns `{:stream_not_found, id}`, that one is a separate change.


whatyouhide (approved):
Nice catch, lovely.

PR #508, opened 2026-09-21
DO NOT MERGE: Apply the acknowledged HTTP/2 header table size to the decoding table

ericmj:
Don't merge this. `mix.exs` points at hpax on GitHub, because `HPAX.protocol_resize/2` is on hpax main (elixir-mint/hpax#32) and isn't released. It needs hpax 1.1.0 published and the dependency changed back to a Hex requirement, `{:hpax, "~> 0.1.1 or ~> 0.2.0 or ~> 1.1"}`, before it can go in.

Based on #507. Retarget to main once that merges.

`put_settings/2` documents and accepts `:header_table_size`, but `@valid_client_settings` didn't list it, so when the server acknowledged the SETTINGS frame the reduce in `apply_client_settings/2` fell through to the catch-all and raised `RuntimeError` out of `stream/2`: "received ack from server for invalid client setting: :header_table_size". The setting is now recorded and the decoding table's permitted maximum follows it.

That maximum is a limit, not the size of the table. The table's own maximum belongs to the server's encoder, which declares it with a dynamic table size update ([RFC 7541 4.2](https://datatracker.ietf.org/doc/html/rfc7541#section-4.2)), and `HPAX.resize/2` moves both. With the server on a 64 byte table holding `:status: 404`, a raise to 8192, and the server then storing `:status: 500`, which evicts 404 from its own table, an indexed reference to entry 63 would decode to the stale 404. With `HPAX.protocol_resize/2` it's `{:compression_error, "unable to decode headers: {:index_not_found, 63}"}`.

`protocol_resize/2` also follows the limit down when it drops below the size the encoder declared. After such a reduction the server has to declare its new size at the start of the next field block, which [RFC 9113 4.3.1](https://www.rfc-editor.org/rfc/rfc9113.html#section-4.3.1) requires the client to treat as a connection error of type COMPRESSION_ERROR if it doesn't, so a block that skips it is `{:compression_error, "unable to decode headers: :missing_size_update"}`.

Four tests: the setting being applied and usable, the two cases above, and a control where the server does declare the reduced size and the response is accepted. The two error cases fail against `HPAX.resize/2`; the other two pass either way.


whatyouhide (approved):
Nice!

